PR #9, opened 2021-02-05
feat(structure): change to the new proposal workflow

terry-xiaoyu:


README.md
```diff
@@ -1,2 +1,13 @@
-# proposals
-EMQ X Enhancement Proposals
+# EMQ X Improvement Proposals (EIP)
```
HJianBo, line 1: Do we need a new name? It looks like the EIP doesn't match the name of our repository - rfc.

HJianBo:
In my opinion, it is not enough to have an active status only? Can we have three states, e.g:

```
Draft -> Accepted or Scheduled? -> Released or Archived?
``` 

1. **Draft** State: Any reasonable request can be merged into this directory. These requests come from within our members or from any type of participant in the community. At this point, the document can roughly be merged in, as long as the request is reasonable enough.

The draft document can be iterated over and over until it looks reasonable and can be implemented. 

2. **Accepted or Scheduled** State: Once the project manager and product manager have validated a draft document and decided to put it into a release, we can move the document from the **Draft** directory to the **Accepted/Scheduled** directory. At this point, we can assign a fixed document number to it.

Documents under Accepted/Scheduled can also be allowed to be modified, but the PR needs to be approved by the project and product manager.

3. **Archived or Released** State: A document is moved to this folder when its content has been completed and verified.

Once a document is in this state, no updates are allowed.

README.md
zmstone: repo has been renamed to eip

zmstone: reasons:
1. RFC does not make much sense (in a general way), it was an arbitrary name which happened to have made its fame
2. we will often refer to a proposal with a number like EIP-1234, having to use RFC-1234 will cause confusion.

zmstone:
> only? Can we have three

we discussed this before, the conclusion is not to turn this into a jira board.

zmstone:
will merge this one for now.
updates are welcome.

PR #14, opened 2021-02-22
feat(design): add 0008-community-plugins

z8674558:
https://github.com/emqx/emqx/issues/3998


design/0008-community-plugins.md
```diff
@@ -0,0 +1,48 @@
...
+
+Allowing people to develop their own plugins is a good way for EMQX to gain popularity.
+To achieve this, it is nice to `approve` some community plugins and let people use them.
+
+## Design
+
+In rebar.config.erl, we can add a list of approved community plugins.
+
+```erlang
+community_deps() ->
```
zmstone, line 24: I'd propose to have this in a separate file e.g. `community-plugins`
and consult it from `rebar.config.erl`

zmstone: the advantage of having it in a separate file is to keep minimum lines of such in `rebar.config.erl` which may otherwise cause more lines of conflicts when porting changes to enterprise.

zmstone:
as we discussed, it is still unclear if we want to keep promoting `plugin` in v4.3 or try to promote `module` in this new framework instead.
in case of `module`, this proposal will
1) depend on a change to be made (merge enterprise to opensource).
2) need updates, e.g. stop using the `plugin` word.

we should also provide a sample plugin (or module) and include it in the proposal
there is a https://github.com/emqx/emqx-plugin-template
and we may need to create emqx-sample-module if to support module instead.

zmstone:
> as we discussed, it is still unclear if we want to keep promoting `plugin` in v4.3 or try to promote `module` in this new framework instead.
> in case of `module`, this proposal will
> 
> 1. depend on a change to be made (merge enterprise to opensource).
> 2. need updates, e.g. stop using the `plugin` word.
> 
> we should also provide a sample plugin (or module) and include it in the proposal
> there is a https://github.com/emqx/emqx-plugin-template
> and we may need to create emqx-sample-module if to support module instead.

We will have to continue supporting plugin, (at least in 4.3)
so please every one, have a look.

design/0008-community-plugins.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+ ## Configuration Changes
+
+
+
+ ## Backwards Compatibility
+
+
+ ## Document Changes
```
zmstone, line 42: we should ensure the new plugin framework is documented somewhere

design/0008-community-plugins.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+
+ ## Testing Suggestions
```
zmstone, line 46: should describe how a third-party plugin can be tested.
ideally the standalone repo should test itself without having to be injected into the umbrella project.
but if there is no way to avoid that, we may need to find a easy way to test (local and in CI) their app.

e.g. put the app under development in emqx's `_checkouts` dir and use emqx's `make ct --dir=...` to test it.

z8674558: > e.g. put the app under development in emqx's _checkouts dir and use emqx's make ct --dir=... to test it.

You expect these steps are done manually? Or there should be some command to pull the repository and place it in a good place to make test works?

z8674558: How about having `make include https://some-plugin.git` command, which will download `some-plugin` to `apps` directory.

zmstone: It's always tricky having to deal with inverse dependency in tests.
ideally, we should provide a `emqx-mock` or something for standalone repo's eunit and ct.
However I realise that could be another EIP.
So as the first step, we can document it as a manual step.
Nothing is stopping anyone from scripting it though.

z8674558: updated the doc

design/0008-community-plugins.md
z8674558: updated the doc

zmstone (approved):

HJianBo (approved):

PR #18, opened 2021-03-03
Create EIP for improved monitoring

k32:


active/0009-improved-monitoring.md
```diff
@@ -0,0 +1,106 @@
...
+                  registered_name = user_drv,stack_size = 10,heap_size = 2586,
+                  total_heap_size = 3196,
+                  current_stacktrace = [{user_drv,server_loop,6,
+                                                  [{file,"user_drv.erl"},{line,191}]}],
+                  group_leader = "<0.152.0>"},
+...
+```
+
+Alternatively, logs can be written in a binary form, to save space.
+Also different kinds of messages can be written to different log files instead of a single one.
```
qzhuyan, line 80: metion which logging system are you going to use?

qzhuyan (approved):

active/0009-improved-monitoring.md
k32: Done, the document mentions kernel logger now

zmstone (approved):

wivwiv:
This is a great function, and the data is an important basis for the correct operation of the system

I think it's better to put it in REST API and StatsD as a supplement to the previous monitoring scheme.

https://github.com/emqx/emqx-prometheus
https://docs.emqx.io/en/broker/latest/advanced/http-api.html#get-api-v4-nodes-node

k32:
@wivwiv 
> I think it's better to put it in REST API and StatsD as a supplement to the previous monitoring scheme.

Thanks for review! REST API definitely should be possible to implement. Do you have any preferences about the schema?
To elaborate on the StatsD proposal... Erlang processes can be short-lived, so reporting them as metrics can lead to dimension explosion. Not all timeseries databases that can handle this kind of load well. Prometheus probably won't be happy with it, the documentations warns against creating too many labels: https://prometheus.io/docs/practices/naming/#labels

HJianBo:
Awsome proposal! I have a small question:

If this is turned on, how much of a performance drain does it have on emqx? My guess is that when there are a lot of processes () in emqx, system_monitor should drain a lot of cpu and memory.

Can we always turn this on if we are in a production environment? If it's possible, this would be really cool

k32:
> If this is turned on, how much of a performance drain does it have on emqx? 

This is something I would like to measure in a simulated load-test with many clients.

> My guess is that when there are a lot of processes () in emqx, system_monitor should drain a lot of cpu and memory.

This is possible. Even though system_monitor's algorithm is `O(n_processes)` in terms of CPU and memory, it has some (configurable) safeguards against running when the number of processes is too large.

> Can we always turn this on if we are in a production environment? 

I have some experience running `system_monitor` in the production environments 24/7. Overall, it never caused any troubles. But the sample size equals 2, so some caution and testing are needed.

HJianBo (approved):

zmstone (approved):

PR #27, opened 2021-03-17
Introduce JQ to emqx rule engine

terry-xiaoyu:


k32:
Great proposal.

active/0011-jq-in-rule-engine.md
```diff
@@ -146,13 +147,67 @@ only output a single map result, and the map will then be piped to the JQ clause
...
 
+The second approach is NIF, with the drawback of more changes to the code (compile
+the code to a dynamic C library rather than a single binary), and safety (it
+brings down the entire erlang system on crash, and may hold up the erlang
+scheduler if it returns too late). But this way has the benefit of efficiency
+and no independent `jq` binary is required.
+
+### A try for using JQ as NIF
+
+An Erlang NIF library has been create at https://github.com/terry-xiaoyu/jqerl.
```
zmstone, line 168: move to https://github.com/emqx/jq

active/0011-jq-in-rule-engine.md
```diff
@@ -146,13 +147,67 @@ only output a single map result, and the map will then be piped to the JQ clause
...
+
+It is now at the very early stage and is not well-tested.
+It requires the `jqlib`, so before using the library you should install the
```
zmstone, line 171: maybe it's in your plan already: compile jqlib source code ?

active/0011-jq-in-rule-engine.md
```diff
@@ -146,13 +147,67 @@ only output a single map result, and the map will then be piped to the JQ clause
...
+`jq` (version 1.6) on your system, and make sure the `libjq.a` can be found
+in `/usr/local/lib`.
+
+The NIF now can only parse JSON strings, the argument `--raw-input` maybe supported
```
zmstone, line 175: you mean as a stream ?

active/0011-jq-in-rule-engine.md
```diff
@@ -146,13 +147,67 @@ only output a single map result, and the map will then be piped to the JQ clause
...
+
+```
+1> jqerl:parse(<<".a">>, <<"{\"a\": ">>).
+{error,{jq_err_parse,<<"Unfinished JSON term at EOF at line 1, column 6 (while parsing '{\"a\": ')">>}}
+```
 
 ## Configuration Changes
 
-No configuration changes.
+If we are about to use erlang port to talk to the independent `jq` command line
```
zmstone, line 207: I believe we can move `jq` command + erlang port to "# Declined Alternatives"

active/0011-jq-in-rule-engine.md
```diff
@@ -146,13 +147,67 @@ only output a single map result, and the map will then be piped to the JQ clause
...
+FreeBSD, Solaris, and Windows for now, so the simplest way is to package the `jq`
+binary along with all of the emqx installation packages, and talk to `jq` using
+the [erlang port](http://erlang.org/doc/tutorial/c_port.html). For someone who
+is building emqx from source code of emqx repo, he can put the jq binary in to
+the right path according to the configuration.
 
+The second approach is NIF, with the drawback of more changes to the code (compile
+the code to a dynamic C library rather than a single binary), and safety (it
+brings down the entire erlang system on crash, and may hold up the erlang
+scheduler if it returns too late). But this way has the benefit of efficiency
```
qzhuyan, line 163: Long-running NIF calls can be done on a dirty scheduler, however, finding the perfect threshold of whether a call should be run on dirty/or none-dirty can be a challenge.

active/0011-jq-in-rule-engine.md
```diff
@@ -146,13 +147,67 @@ only output a single map result, and the map will then be piped to the JQ clause
...
 I am happy with this syntax now.
 
-### Introduce JQ as NIF
+### Introduce JQ as Port or NIF
 
+JQ is written in portable C as a single binary, reading command line argument
+from stdin and outputting results to the stdout. It supports Linux, OS X,
+FreeBSD, Solaris, and Windows for now, so the simplest way is to package the `jq`
+binary along with all of the emqx installation packages, and talk to `jq` using
+the [erlang port](http://erlang.org/doc/tutorial/c_port.html). For someone who
```
qzhuyan, line 156: emqx probably needs multi port programs in order to make things run in parallel?

active/0011-jq-in-rule-engine.md
qzhuyan: 👍

active/0011-jq-in-rule-engine.md
terry-xiaoyu: Yes, let user install jq before building/installing emqx is not a good idea. It would lead to some problem if the use installed the wrong jq version.

I think we should also compile our `libcrypto`, as it is a common problem that emqx fails to start due to the wrong ssl version installed in the system.

active/0011-jq-in-rule-engine.md
terry-xiaoyu: Not streaming, I don't want to support the option `--stream` as I suppose there would't be many user ask it.
But the `--raw-input` option maybe useful to process `plain text`:

> --raw-input/-R:
> Don't parse the input as JSON. Instead, each line of text is passed to the filter as a string. If combined with --slurp, then the entire input is passed to the filter as a single long string.
> 
e.g. 
To split a plain text into an array and remove the last empty `\n`:
```
echo "1\n2\n3" | jq --slurp --raw-input 'split("\n")[:-1]'
```

active/0011-jq-in-rule-engine.md
terry-xiaoyu: That's good.

active/0011-jq-in-rule-engine.md
terry-xiaoyu: It can be a hard work and I may need your help :)

I can split the large JSON to several pieces of text and call `jq:parse/2` multiple times. I will try it to see if it brings some benefits.

active/0011-jq-in-rule-engine.md
terry-xiaoyu: Yes, we need to start multiple ports, each for a rule (added by emqx rule engine).

active/0011-jq-in-rule-engine.md
terry-xiaoyu: I will, after you give me the write permission to the repo. ;-)

active/0011-jq-in-rule-engine.md
zmstone: maybe more feasible to statically link crypto when building otp.
ref:  http://erlang.2086793.n4.nabble.com/static-build-of-Erlang-with-openssl-td4681642.html

active/0011-jq-in-rule-engine.md
```diff
@@ -0,0 +1,344 @@
...
+
+If there's some error in the jq filter or the json string, `{error, Reason}` will
+be returned:
+
+```
+1> jqerl:parse(<<".a">>, <<"{\"a\": ">>).
+{error,{jq_err_parse,<<"Unfinished JSON term at EOF at line 1, column 6 (while parsing '{\"a\": ')">>}}
+```
+
+### Update 2022-06-09:
```
zmstone, line 278: please update the original text. I don’t think we need to keep all history revisions.

active/0011-jq-in-rule-engine.md
```diff
@@ -0,0 +1,344 @@
...
+interface supports both a NIF based backend and port based backed. The user can
+configure which backend to use. The port based one is safer (as it cannot crash
+the Erlang VM and cannot cause memory leaks and memory fragmentation inside the
+VM) but the NIF backend is faster. At the time of writing, the JQ function in
+the rule engine can only use the port based backend as this backend is
+currently the only one that supports timeouts. There is a plan to also support
+timeouts in the NIF based backend. When that is possible, EMQX users should be
+given the option to configure which backend to use.
+
+## Configuration Changes
```
zmstone, line 291: This has been covered in the previous section where the timeout is discussed

active/0011-jq-in-rule-engine.md
kjellwinblad: Ok, will do so and will remove the other "Update" subsections to their parent section

active/0011-jq-in-rule-engine.md
kjellwinblad: > I can split the large JSON to several pieces of text and call jq:parse/2 multiple times. I will try it to see if it brings some benefits.

I think that could help in some cases but it would not be a general solution as JQ can spend an infinite amount of time on a small piece of JSON.

active/0011-jq-in-rule-engine.md
kjellwinblad: As you know, the current implementation use a fixed number of long running port porgrams.

PR #28, opened 2021-03-18
(0002-new-config-syntax) add implementation, delete yaml

z8674558:


k32:
Can we perhaps consider https://dhall-lang.org/ as an alternative? It's fairly advanced (it has types, functions and semantic diff), but it can be transpiled to yaml or json.

k32 (approved):

PR #32, opened 2021-05-24
feat(0012): better authentication

tigercl:


zmstone:
This doc lack of the API design proposal.

what's the config look like if chains are to be provisioned from files.
what are the APIs to support dynamically compose an auth chain.

k32 (approved):

PR #39, opened 2021-08-11
cluster call for  the consistency of resources

zhongwencool:


active/0012-cluster-call.md
```diff
@@ -0,0 +1,168 @@
...
+## Design
+
+This proposal uses mnesia to record the execution status of MFA, to ensure the consistency of the final resources & data in the cluster.
+
+This proposal is not applicable to high frequency request calls.
+
+### mnesia table structure
+
+```erlang
+-record(cluster_rpc, {tnx_id :: pos_integer(), mfa :: mfa(), pending_nodes = [] :: [node()]}).
```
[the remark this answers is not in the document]
zhongwencool: Good Point!

active/0012-cluster-call.md
```diff
@@ -0,0 +1,179 @@
...
+## Design
+
+This proposal uses mnesia to record the execution status of MFA, to ensure the consistency of the final resources & data in the cluster.
+
+This proposal is not applicable to high frequency request calls, all updates are performed in strict order.
+
+### mnesia table structure
+
+```erlang
+-record(cluster_rpc, {tnx_id :: pos_integer(), mfa :: mfa()}).
```
zmstone, line 33: maybe add two more fields for troubleshooting & logging. `timestamp`, and `initiator` (the node).

active/0012-cluster-call.md
```diff
@@ -0,0 +1,179 @@
...
+                        TnxId = incr_tnx_id(),
+                        mnesia:write(#cluster_rpc{tnx_id = TnxId, mfa = MFA}),
+                        %% check new node join in but execute none transaction yet.
+                        [begin
+                             case mnesia:read(emqx_cluster_tnx_id, N) of
+                                 [] -> mnesia:write(#cluster_tnx_id{node = N, tnx_id = TnxId - 1});
+                                 _ -> ok
+                             end
+                         end || N <- ekka:nodelist()],
+                        mnesia:write(#cluster_tnx_id{node = Node, tnx_id = TnxId}),
```
zmstone, line 83: my thoughts about the usage of `emqx_cluster_tnx_id`:
This table is used to keep track of the last applied transaction id of their OWN.
The transaction initiator should NOT write records for other nodes.

So there is no need to lock `emqx_cluster_tnx_id` table in the beginning of the transaction.

active/0012-cluster-call.md
```diff
@@ -0,0 +1,201 @@
...
+           retry -> {error, "catch up failed"};
+           ok ->
+               Trans = fun() -> apply_to_commit(Node, MFA) end,
+               transaction(Trans)
+       end.
+   apply_to_commit(Node, MFA) ->
+       mnesia:write_lock_table(?CLUSTER_MFA),
+       #cluster_call_cursor{tnx_id = LastDoneId} = mnesia:read(?CLUSTER_CURSOR, Node),
+       #cluster_call_mfa{tnx_id = LatestId} = mnesia:last(?CLUSTER_MFA),
+       ok = do_catch_up_in_one_trans(LatestId, LastDoneId, Node),
```
zmstone, line 136: we need some comment here to explain why there should be a catch up again after line 126.

active/0012-cluster-call.md
```diff
@@ -0,0 +1,239 @@
...
+Update resources may be lost during the RPC call.
+
+- If there is a network disturbance during RPC, it may cause the RPC to fail.
+- If a remote operation to update a resource fails, there is no retry or any other remedy, causing inconsistent configuration in the cluster.
+- If multiple updates are performed concurrently, it may happen that node 1 performs updates in order 1, 2, 3, but node 2 updates in order 1, 3, 2. There is no order guarantee.
+- Lack of replay. If a node is down for a while, there is a lack of history event replay to catch up with the changes happened during the down time.
+
+## Design
+
+This proposal uses mnesia to record the execution status of MFA, to ensure the consistency of the final resources & data in the cluster.
```
k32, line 26: If the table contains only ops, how to provision a new node?

k32 (approved):
LGTM, but I would like to see more testing scenarios. Plan for some fault injection tests, perhaps?

qzhuyan:
I think rpc and mnesia transaction are correlated, if rpc call/cast is failing due to network issue, mneisa transaction will be aborted as well.

active/0012-cluster-call.md
```diff
@@ -0,0 +1,239 @@
...
+   get_next_mfa(Node) ->
+       Fun =
+           fun() ->
+               case mnesia:read(?CLUSTER_CURSOR, Node) of
+                   [] ->
+                       LatestId = get_latest_id(),
+                       mnesia:write(#cluster_rpc_cursor{node = Node, tnx_id = LatestId}),
+                       caught_up;
+                   [#cluster_rpc_cursor{tnx_id = DoneTnxId}] ->
+                       case mnesia:read(?CLUSTER_MFA, DoneTnxId + 1) of
```
qzhuyan, line 98: the `read` call here needs to take a write lock.

active/0012-cluster-call.md
```diff
@@ -0,0 +1,239 @@
...
+                       end;
+                   _ -> {next_state, ?CATCH_UP, Data, [?CATCH_UP_AFTER(1)]}
+               end;
+           {aborted, _Reason} -> {next_state, ?CATCH_UP, Data, [?CATCH_UP_AFTER(1)]}
+       end.
+   
+   get_next_mfa(Node) ->
+       Fun =
+           fun() ->
+               case mnesia:read(?CLUSTER_CURSOR, Node) of
```
qzhuyan, line 92: the `read` call here needs to take a write lock.

active/0012-cluster-call.md
```diff
@@ -0,0 +1,239 @@
...
+       case do_catch_up(LatestId, Node) of
+           caught_up -> ok;
+           ok -> do_catch_up_in_one_trans(LatestId, Node);
+           _ -> mnesia:abort("catch up failed")
+       end.
+   ```
+
+   **Risk point**: If the previous unfinished MFA in the `init_mfa` transaction is executed successfully, but the latest MFA fails and leads to abort,  it will roll back the previous unfinished MFA as well, thus causing the MFA to be executed again later. So MFA must be idempotent.
+
+   If nodes A,B have completed transaction 4, and they are fighting to update transaction 5 at the same time, A finally gets transaction 5 and commits successfully, at this time, B can only update to transaction 6, when the event of transaction 5 has not reached node B, B got transaction 6, it will lead to the error that transaction 6 is executed before transaction 5. So we must check  if there are any uncompleted transactions in transaction 6(`do_catch_up_in_one_trans`).
```
qzhuyan, line 186: Can mnesia `mnesia:sync_transcation` solve the race here?

active/0012-cluster-call.md
```diff
@@ -0,0 +1,201 @@
...
+   do_catch_up_in_one_trans(LatestId, LatestId, _Node) -> ok;
+   do_catch_up_in_one_trans(LatestId, LastDoneId, Node)when LatestId > LastDoneId ->
+       ToTnxId = LastDoneId + 1,
+       do_catch_up(Node, ToTnxId),
+       do_catch_up_in_one_trans(Node, LatestId, ToTnxId).  
+   ```
+
+   **Risk point**: If the previous unfinished MFA in the `apply_to_commit` transaction is executed successfully, but the latest MFA fails and leads to abort,  it will roll back the previous unfinished MFA as well, thus causing the MFA to be executed again later. So MFA must be idempotent.
+
+5. Because all ekka's transactions execute on core node, but the MFA should execute on all node. So we must do MFA with RPC:
```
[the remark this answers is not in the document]
zmstone: replicant nodes can not write initiate transactions, it RPCs a core node to update its last-applied ID to itself.
so it should first, try to apply MFA locally, then commit its last-applied ID in the database via ^ this flow

zmstone:
> I think rpc and mnesia transaction are correlated, if rpc call/cast is failing due to network issue, mneisa transaction will be aborted as well.

The idea is to make MFA idempotent,
Two reasons:
1. the retry-nature of mnesia transaction functions
2. the lack of rollback for the MFAs

zmstone:
> > I think rpc and mnesia transaction are correlated, if rpc call/cast is failing due to network issue, mneisa transaction will be aborted as well.
> 
> The idea is to make MFA idempotent,
> Two reasons:
> 
> 1. the retry-nature of mnesia transaction functions
> 2. the lack of rollback for the MFAs

that's why MFA can be, and should be applied *before* committing the last-applied ID to database.

PR #40, opened 2021-08-12
Different Authentication Methods on Zones

1995parham:
Based on [discussion](https://github.com/emqx/emqx/discussions/5459)

### DIFF
--- a/active/0012-zone-based-auth.md
+++ b/active/0012-zone-based-auth.md
@@ -8,7 +8,7 @@ It should be possible to define multiple zones with different authentication sta
 
 ## Motivation
 
-We have two types of clients, one of them connects using external network and needs to be authenticated by http
+We have two types of clients, one of them connects through an external network and needs to be authenticated by http,
 but another group connects using an internal network and we need only jwt authentication for them.
 
 ## Design
--- a/active/0012-zone-based-auth.md
+++ b/active/0012-zone-based-auth.md
@@ -4,7 +4,7 @@
 
 ## Abstract
 
-We can define multiple zones with multiple authentication stack on emqx.
+It should be possible to define multiple zones with different authentication stacks.
 
 ## Motivation
 
--- a/active/0012-zone-based-auth.md
+++ b/active/0012-zone-based-auth.md
@@ -15,7 +15,51 @@ but another group connects using an internal network and we need only jwt authen
 
 ## Configuration Changes
 
-There should be zone specific list of authentications.
+There should be zone specific list of authentications which has more priority than global rule and override them.
+
+```
+authz: {
+    rules: [
+       {
+           type: redis
+           config: {
+              servers: "127.0.0.1:6379"
+              database: 0
+              pool_size: 1
+              password: public
+              auto_reconnect: true
+              ssl: {enable: false}
+           }
+           cmd: "HGETALL mqtt_authz:%u"
+       }
+    ]
+}
+
+zone.x {
+  authz: {
+      rules: [
+         {
+             type: mysql
+             config: {
+                server: "127.0.0.1:3306"
+                database: mqtt
+                pool_size: 1
+                username: root
+                password: public
+                auto_reconnect: true
+                ssl: {
+                  enable: true
+                  cacertfile:  "etc/certs/cacert.pem"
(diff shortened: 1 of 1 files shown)


active/0012-zone-based-auth.md
```diff
@@ -0,0 +1,30 @@
+# Authentication based on Zone
+
+## Changelog
+
+## Abstract
+
+We can define multiple zones with multiple authentication stack on emqx.
```
k32, line 7:
```suggestion
It should be possible to define multiple zones with different authentication stacks on emqx.
```

active/0012-zone-based-auth.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Motivation
+
+We have two types of clients, one of them connects using external network and needs to be authenticated by http
```
k32, line 11:
```suggestion
We have two types of clients, one of them connects through an external network and needs to be authenticated by http,
```

active/0012-zone-based-auth.md
```diff
@@ -0,0 +1,30 @@
...
+but another group connects using an internal network and we need only jwt authentication for them.
+
+## Design
+
+## Configuration Changes
+
+There should be zone specific list of authentications.
```
k32, line 18: Could you clarify what the configuration should look like? In 5.0 we'll have the following hierarchy: 
```hocon
authorization:{
    rules: [
    ]
}
...
zone.x {
  a: {b:1, c: 1}
  listeners.y {
     a: {b: 2}
   }
}
```
Should we use `authentication` subtree as a default, with the possibility of adding zone-specific overrides?

1995parham: Yes, exactly. we can have exact same subtree under any zone.

k32: Could you please add the example config to the EIP?

1995parham: Yes, sure I will add it now.

1995parham: I have update it based on [emqx_authz](https://github.com/emqx/emqx/tree/master/apps/emqx_authz). Also, the new configuration is really great but I didn't have a chance to use it.

k32: Thanks, now the EIP looks clear to me.

k32:
LGTM, let's wait what my colleagues who worked on the enhanced authentication think about this.

zmstone:
Hi. Thank you for the proposal.

We are still iterating on the authn (authentication) and authz (authorization, previously ACL) implementations quite a bit,
we have recently decided to make some adjustments in configuration layout.
to summarise at very high level.
1. previously (in alpha releases) listeners were designed too deep in the config (as sub-struct of a zone), it will be moved to up as a root level config
2. authentication config will be moved down to listener (embedded as a sub-struct, or using a reference is still in debate)
3. authorization is a singleton, meaning there is only going to be one instance of authorization rule set. 

elaborating more in the changes for authn and authz:

`authn` is quite coupled with access endpoint and protocol (listeners), hence not suitable to use a generic authn instance for all kinds of different accesses. e.g. SSL listener may have certificate based authn, but not plaintext TCP.

`authz` has its scope in the entire routing table (topics), that is, no matter from which listener or zone the client is connected, they see the routing table equally.

We understand the requirement of having different rule-sets for internal / external though, an alternative is to have the zone (and maybe even user properties in the future) to be parameterised for rules.

 in file based rules, we currently support filtering on auth principals.
e.g.
```
{ principal: {"and" : [{username: "^test?", clientid: "^test?"}]}
  topics: [test1, test2],
  action: publish
  permission: allow
}
```
if we want to allow clients authenticated to access internal zone to publish to any topic, we may add zone as a auth principal like below.
```
{  principal: {zone => internal}}
   topics: ["#"],
   action: publish
   permission: allow
}
```

for db-backend based authz checks, we can add a `${zone}` place holder in the queries.

zmstone:
ping @Rory-Z for comments

Rory-Z:
`principal: {zone: internal}}` is good idea, we are still having a heated discussion about `authz`, I will update later for more details

zmstone:
5.0 implemented per-listener authentication.

PR #46, opened 2021-09-13
feat(0013): Add initial version of configuration migration EIP

k32:


active/0013-configuration-migration.md
```diff
@@ -0,0 +1,141 @@
...
+* 2021-09-09: @k32 Initial draft
+
+## Abstract
+
+Backup and restore of database tables and configuration should be used for the cluster upgrade.
+This procedure should replace the JSON configuration dumps.
+
+## Motivation
+
+Upgrading the EMQ X cluster between major releases requires redeployment of the cluster.
```
zmstone, line 14: actually minor releases too.

active/0013-configuration-migration.md
```diff
@@ -0,0 +1,141 @@
...
+https://docs.emqx.io/en/broker/v4.3/advanced/data-import-and-export.html#data-import-and-export
+
+Encoding Erlang terms to JSON and decoding them back is a manual error-prone process.
+We propose to use binary dumps instead, starting from EMQ X 5.0.
+
+## Design
+
+A new `emqx_metadata` mnesia table should be added.
+Mnesia has a checkpoint feature that can be used to perform a backup.
+TODO: decide the best way to include data from the external databases (MySQL, MongoDB, etc.) to the backup.
```
zmstone, line 25: that's probably out of the scope of this proposal

zmstone:
It's great that you proposed to handle hop beam upgrade cases, which is a 'nice to have' feature.
I'd propose to confine the scope with only table schema transformation.

During patch releases:
For patch releases, we should avoid doing table schema changes, and for data changes, code will have to be backward compatible, i.e. no need to do table transformation.
For configs, we'll have to write backward compatible schema (and perhaps code which makes use of the configs). i.e. no need to do config transformation 

to sum up, the requirement is straightforward:
provide a script which can:
   * dump the database to a file, and also collect config files (optional, e.g. for k8s deployment, configs are usually managed outside)
   * transform data (maybe schema too) to make it compatible with the new release

k32:
@zmstone Thanks for the review, applied remarks

active/0013-configuration-migration.md
```diff
@@ -0,0 +1,162 @@
...
+There should be an Erlang module called `emqx_upgrade` that collects all upgrade hooks.
+
+```erlang
+-module(emqx_upgrade).
+
+-export([upgrade_hooks/0]).
+
+upgrade_hooks() ->
+   %% Num | Intruduced in release | Function
+    [ {100, "5.1.0",                fun upgrade_lib_baz/2}
```
zmstone, line 67: if this is to be compared, maybe a tuple instead?
i.e. `"5.1.0"` -> `{5, 1, 0}`

active/0013-configuration-migration.md
```diff
@@ -0,0 +1,162 @@
...
+1. A strictly decreasing number of the upgrade hook
+1. Version of EMQ X release where this upgrade hook has been introduced, non-increasing
+1. Reference to the upgrade hook.
+   All the upgrade hooks should be immutable: once they are written, they shouldn't change
+
+There should be a separate module that works with the upgrade hooks, called `emqx_upgrade_helper`.
+
+It has the following functions:
+
+1. `init()`, where it creates an internal metadata table that contains `data_version` record that contains the EMQ X release version corresponding to the currently loaded configuration and runtime data.
```
zmstone, line 88:
```suggestion
1. `init()`, where it creates the internal metadata table `emqx_metadata` which contains `data_version` record that contains the EMQ X release version corresponding to the currently loaded configuration and runtime data.
```

active/0013-configuration-migration.md
```diff
@@ -0,0 +1,162 @@
...
+   1. Execute the hooks obtained on the previous step.
+      If the hook ran successfully, update `last_upgrade_hook` counter.
+      If `introduced_in` field of the _next_ update hook is larger than the current version, then insert the _current_ version to `emqx_metadata` table
+
+#### REST API
+
+Management API should be extended with two methods:
+
+```bash
+$ curl -i --basic -u admin:public -X POST "http://localhost:8081/api/v4/data/backup"
```
zmstone, line 108:
```suggestion
$ curl -i --basic -u admin:public -X POST "http://localhost:8081/api/v5/data/backup"
```

zmstone:
For 5.0 to 5.1 upgrade, we may want to perform rolling upgrade for the cluster instead of upgrading the nodes into a different cluster.
To meet this requirement, I wonder if it's possible at all to change Mnesia schema the way in this proposal.

zmstone:
> For 5.0 to 5.1 upgrade, we may want to perform rolling upgrade for the cluster instead of upgrading the nodes into a different cluster.
> To meet this requirement, I wonder if it's possible at all to change Mnesia schema the way in this proposal.

1. if we do `mnesia:transform_table` on the fly, do we still need the BUP (we may need BUP for backup/restore, but not necessary for upgrade purposes)?
2. `mnesia:transform_table` probably enables schema change for the `core` nodes, what about the replicant nodes ?

PR #48, opened 2021-09-28
fix(0014): Extend the proposal

k32:


active/0014-rolling-cluster-upgrade.md
```diff
@@ -53,6 +66,13 @@ There are two major areas that need to be considered to support this kind of upg
 1. RPC compatibility
 1. Mnesia schema backward-compatibility
 
+### Design of the helper bpapi application
+
+We propose to create a new helper application that helps to maintain the API backward-compatibility.
```
zmstone, line 71: Is this application generic enough to be used in any other Erlang project ?
if it's very much tied with emqx, maybe put it in `apps/emqx/src/bpapi`

external plugins will need to add emqx (sub_dir) as a dependency anyway.

k32: My initial thought was that we may want to use it in Mria, for example... But perhaps it's an overkill, since we can just modify it to never try to connect to the upstream nodes running a different version.

active/0014-rolling-cluster-upgrade.md
```diff
@@ -75,31 +95,48 @@ Non-trivial changes to the schema should be performed in stages:
...
+    ...
 ```
 
 The following static checks should be written:
 
 1. (xref): Only the functions in `*_bpapi` modules, that have `-introduced_in()` annotation, are allowed to call `rpc:call` and `gen_rpc:call` function
-1. Functions that have `deprecated_in` annotation are not removed earlier than two releases after the deprecation.
+1. Functions that have `deprecated_in` annotation are not removed earlier than two releases after the deprecation
+1. Each API function does an RPC to an existing function (xref check)
+1. Compatibility of the typespecs of the API modules:
```
k32, line 135: Now I realize that this check is less useful than it looks, because it only verifies that an old version of the code can call the new version of the code, but not the other way around.

active/0014-rolling-cluster-upgrade.md
zmstone: ok. let's do it as a part of emqx app, and split it out if it turn out to be useful as a standalone.

active/0014-rolling-cluster-upgrade.md
```diff
@@ -75,31 +95,48 @@ Non-trivial changes to the schema should be performed in stages:
...
+   - Function domain is not reduced
```
zmstone, line 136: what's a domain ?

k32: It's roughly the set of input values that the function accepts.
However, now I am thinking that this check will be rather useless.

zmstone (approved):

PR #49, opened 2021-10-05
feat(0015): move authn doc here

zmstone:


tigercl (approved):

active/0015-unified-authentication.md
```diff
@@ -0,0 +1,169 @@
...
+```
+POST /listeners/:listener-name/authentication
+GET /listeners/:listener-name/authentication
+GET /listeners/:listener-name/authentication/:id
+DELETE /listeners/:listener-name/authentication/:id
+PUT /listeners/:listener-name/authentication/:id
+PATCH /listeners/:listener-name/authentication/:id
+```
+
+A listener name is of foramt `protocol:id` which is assigend in the config file, e.g.
```
qzhuyan, line 153: s/foramt/format

active/0015-unified-authentication.md
```diff
@@ -0,0 +1,169 @@
...
+
+We should allow users to compose (configure) a "chain" of backends with a defined order in which the
+checks are performed one after another. Each check against the backend in the chain may yield 3 different
+results for one-request authentication:
+
+- `ignore` is to indicate there is no auth information found hence should
+   continue validate the client against the rest of the backends in the chain.
+- `{ok, Info}` as a login accepted, hence to terminate the auth calls from here,
+   where `Info` may contain additional information such as to indicate if the user is a super-user.
+- `{error, Reason}` to indicate that client's login should be denied.
```
qzhuyan, line 66: nit, maybe it is due to backward backward compability,
I prefer more understandable ret values such like
- accept
- {reject, Reason}
- contitune 
- {error, Reason} %% when backend has other issues.

And also a default policy should be set on the chain such like `reject` or `accept`

qzhuyan (approved):

active/0015-unified-authentication.md
zmstone: there is never going to be a default `accept`, if all auth backends `ignore`, the result is `not_authorized`.

I agree `accept` `reject` and `continue` are better indicators, @savonarola @tigercl please consider a refactoring.

active/0015-unified-authentication.md
```diff
@@ -0,0 +1,174 @@
...
+  user_id_type: clientid
+}
+```
+
+- Example config for built-in-database (mnesia) username/password based per-listener auth
+
+```
+listener.tcp.default {
+  ...
+  authentication: {
```
savonarola, line 108: In this example and the example above we have single authenticator config as `authentication` value; but due to the proposal we may have a chain of authenticators, shouldn't we have a list here?

savonarola (approved):

PR #50, opened 2021-10-12
introduce emqx_conf application.

zhongwencool:


active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+- This file can only be modified via the API and manual modification of file directly is not supported.
+
+- When the configuration of a specific node is updated via HTTP API, it will be persisted to this file.
+
+#### emqx_conf application
+Before, the initialization of the configuration file, cluster_call, was done through the `emqx_machine`, we will split this part of the functionality from the` emqx_machine` and make a new application `emqx_conf`
+
+The role of this application is to
+- Converting the configuration from hocon format to erlang sys.config format at initialization.
+- Manage hot updates and deletions of configurations within the cluster.
```
zmstone, line 58:
```suggestion
- Manage live-updates and deletions of the configurations, and replicate across the cluster
```

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+- Maintain data consistency for the same configuration, keeping the same configuration in the cluster must be the same, such as `authorization.source`, rules engine's rules.
+- Different nodes of the cluster may set different values for the same configuration item, such as: use different `rate_limit` settings for machines with different hardware limitations.
+
+## Design
+
+### Configuration files
+
+#### emqx.conf
+
+Emqx reads `emqx.conf` forconverting this hocon file into Erlang format at startup, and `emqx.conf` has only 2 lines by default.
```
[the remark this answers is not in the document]
zmstone: My concern about "only 2 lines"
it is probably going to be a quite large task.
I would recommend NOT to scope it in this EIP.

We can come back and update this one as step 2,
or create another EIP to focus on that change.

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+- When the configuration of a specific node is updated via HTTP API, it will be persisted to this file.
+
+#### emqx_conf application
+Before, the initialization of the configuration file, cluster_call, was done through the `emqx_machine`, we will split this part of the functionality from the` emqx_machine` and make a new application `emqx_conf`
+
+The role of this application is to
+- Converting the configuration from hocon format to erlang sys.config format at initialization.
+- Manage hot updates and deletions of configurations within the cluster.
+
+The basic node configuration hot update remains in emqx, in other words: `emqx_config_handler`, remains unchanged, but other apps that want to update the configuration must call through the `emqx_conf`'s' API, which cannot call emqx API directly.
```
zmstone, line 60: This is a new feature for 5.0, so we are comparing it to 4.3, not 5.0 beta.1
Describing the changes on beta.1 will soon become stale.
So we should introduce `emqx_config_handler` as a new module.

It should be enough just to elaborate a bit on the layering of the APIs (line 64)

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+### HTTP API design
+
+#### Get the whole configurations.
+
+```erlang
+#{
+  get => #{
+    description => <<"Get all the configurations of cluster, including hot and non-hot updatable items.">>,
```
zmstone, line 74:
```suggestion
    description => <<"Get all the configurations of a given node, or all nodes in the cluster.>>,
```

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+    parameters => [
+       {node, hoconsc:mk(typerefl:atom(),
+          #{in => query, required => false, example => <<"emqx@127.0.0.1">>,
+          desc => <<"Node's name: If you have this field defaulted, the whole cluster is return.">>})},
```
zmstone, line 78:
```suggestion
          desc => <<"Node  name. When this parameter is not provided, configs for all nodes in the cluster are returned">>})},
```

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+       {debug, hoconsc:mk(typerefl:boolean(), #{in => query, required => false,
+          desc => <<"Carries debug(meta data) information, such as:line number,default value,documentation">>})}],
```
zmstone, line 80: I am not sure how easily fill in default value and documentation into metadata.
Maybe leave this feature out for now.

```suggestion
          desc => <<"Carries debug (metadata) information, such as file name and line number">>})}],
```

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+    }.
+```
+
+- get specific configuation, such as: `/configs/emqx_dashboard` will return :
+
+  ```erlang
+  #{'emqx@127.0.0.1' => 
+     #{default_password => "public",
```
zmstone, line 123: There should be a `sensitive` flag in the schema for this field, and the value should be reported back as `"******"` in the API.

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+       listeners =>
+         [#{backlog => 512,inet6 => false,ipv6_v6only => false,
+            max_connections => 512,num_acceptors => 4,port => 18083,
+            protocol => http,send_timeout => 5000}],
+       sample_interval => 10,token_expired_time => 3600000},
+    'emqx1@127.0.0.1' => ...
+  }   
+  ```
+
+- Update specific configuation without query string, will modify the configuration of all nodes in the cluster and persist it in `emqx_cluster_override.conf`. In other words, it is the configuration that keeps the cluster strongly consistent.
```
zmstone, line 134:
```suggestion
- Update specific configuation without the 'node' query string will modify the configuration of all nodes in the cluster and persist it in `emqx_cluster_override.conf`.
```

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+
+The basic node configuration hot update remains in emqx, in other words: `emqx_config_handler`, remains unchanged, but other apps that want to update the configuration must call through the `emqx_conf`'s' API, which cannot call emqx API directly.
+The specific flow is: 
+
+```bash
+Other Apps(eg: emqx_resource) => emqx_conf => emqx API.
+```
+
+### HTTP API design
+
```
qzhuyan, line 68: Do we need a API for `validate` the config before apply?

qzhuyan:
FYI, (If we are planning big)
I just borrow some ideas from `netconf` and do some adjustments:

If one node is able to handle following APIs, a cluser wide config management will be easily achieved with strong consistent , rollback on error, backup/restore config on files.
- `get` : return both read/write config
- `get-config` :  return writable config 
- `edit-config`: create/change config
- `delete-config`: delete part of the config.
- `merge-config`: merge a subtree of config, as consequence there will be lots of subsequent API calls. This is the hard part. 
- `lock`: lock for config on the node, no writes from other O&M system 
- `unlock` :  unlock the config on the node 
- `restore` :  completely reload the config
- `commit`: commit the modified changes and then calling the side effects.

active/0016-emqx-conf.md
zhongwencool: ok, file name and line number first come.

active/0016-emqx-conf.md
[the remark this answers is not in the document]
zhongwencool: Create a new EIP should be better.
I update:
...
and `emqx.conf` explicitly include `override.conf` at the end of the file.
....
For now, we need to integrate all the configurations into `emqx.conf`. But it is planned that `emqx.conf` will no longer contain configurations that use default values in the next phase(propose another EIP), the default is only embedded in the code. Also, users can view all configuration items via the HTTP API (described later). This would have 2 benefits: 

- In subsequent version upgrades, adding/removing/updating configurations will not be overwritten by `emqx.conf`. 
- This allows the user to focus only on the configuration that has been modified.

active/0016-emqx-conf.md
```diff
@@ -0,0 +1,184 @@
...
+### Configuration files
+
+#### emqx.conf
+
+Emqx reads `emqx.conf` forconverting this hocon file into Erlang format at startup, and `emqx.conf` has only 2 lines by default.
+
+```erlang
+include "./data/emqx_cluster_override.conf"
+include "./data/emqx_local_override.conf"
```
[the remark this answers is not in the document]
zhongwencool: https://github.com/emqx/hocon/pull/147

PR #67, opened 2022-07-25
0021: Persistent sessions

ieQu1:


active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
...
+Suppose every topic level takes two bits in the resulting hash and `N_l` equals 3.
+Suppose topic index of this topic filter is `101001`.
+We can create a the bitmask corresponding to this topic filter by setting bits corresponding to the wildcard levels to zeros: `110011`.
+
+Therefore iterating the database can be done using the following pseudocode:
+
+```erlang
+replay(TopicFilter) ->
+  Bitmask = gen_bitmask(TopicFilter),
+  TopicIndex = gen_topic_index(TopicFilter) band Bitmask,
```
thalesmg, line 57: Q: shouldn't the bitmask already be part of `gen_topic_index`?

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
...
+  It = db:seek(TopicIndex),
+  replay(TopicFilter, TopicIndex, Bitmask, It).
+
+replay(TopicFilter, TopicIndex, Bitmask, It) ->
+  {Key, NextIt} = db.next(It),
+  {KTopicIndex, KTimestamp} = parse_key(Key),
```
thalesmg, line 63: The clientid is also part of the key?

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
...
+  if TopicIndex bxor (KTopicIndex band Bitmask) =:= 0 ->
```
thalesmg, line 64: Didn't quite understand this xor operation :see_no_evil: 

Why do we take the xor of those values?  

Apparently, it'll only be `0` when `TopicIndex` has its ones equal to the masked `KTopicIndex` ones.  But, between values of the iterator, won't there be values that give a non-zero xor result? 

For example:

```
...
101001
101010
101011
...
```
?

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
...
+While this is ideal for optimizing DB access while reading historical data, it leaves a significant problem:
+realtime data is not added to the DB in historical order.
+So the client replaying messages that match topics `foo/1` and `foo/2` will no longer see any messages added to `foo/1`, once it starts replaying the keys corresponding to `foo/2`.
+
+Below is the sketch of the solution (thorough testing and modelling is needed):
+
+1. Client connects with clean start =:= `false`.
+1. A client handler process is started.
+1. The client handler process initiates message replay for the topic filters from the saved session in separate processes.
+1. Once the replay reaches a next topic index (i.e. when it finishes replaying messages for index `101010` and starts replaying `101011`, it sends a message to the client handler with the last key that has `101010` prefix
```
thalesmg, line 102:
```suggestion
1. Once the replay reaches a next topic index (i.e. when it finishes replaying messages for index `101010` and starts replaying `101011`), it sends a message to the client handler with the last key that has `101010` prefix.
```

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
...
+1. Client handler process subscribes to the realtime flow of the messages in the topics matching this index, the messages are buffered
```
thalesmg, line 103: Which cilent handler?  The one that was handling `101010`?

From the previous item, it seems that there are multiple client handlers at play.

active/0021-rocksdb-message-persistence.md
thalesmg: If the replay is done serially by topic, won't there still be a problem of replaying the messages in historical order?

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
...
+- Replication strategy is easier to reason about
+- Replaying is easier to implement
+
+Cons:
+- Load may become unbalanced.
+  Some topics can be see significantly more load than the others, particularly in fan-in scenario.
+
+- More backplane network transfers will be involved, since the LB is not aware of the topics that the clients publish to.
+  Messages have to be forwarded through backplane network to the appropriate node.
+  This disadvantage is a relatively insignificant when replication factor is high.
```
thalesmg, line 127:
```suggestion
  This disadvantage is relatively insignificant when replication factor is high.
```

active/0021-rocksdb-message-persistence.md
ieQu1: This is meant to illustrate that certain bits of TopicIndex are zeroed.

active/0021-rocksdb-message-persistence.md
ieQu1: In this example I omitted many irrelevant parameters.

active/0021-rocksdb-message-persistence.md
thalesmg: While looking at how RocksDB does the range iterations, it seems that the most it can do is a prefix scan. So, if we want to iterate over the middle segment of the topic as in the example (`a/+/c`), we'd still iterate over all the present third segments.

Maybe we'll have to do the multiple indexing as Ilya has done for retainer to arrange the topics so that a prefix scan can be used?

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
+# An Example of EMQ X Improvement Proposal
```
thalesmg, line 1: Change title?

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
...
+1. Client handler process subscribes to the realtime flow of the messages in the topics matching this index, the messages are buffered
+1. It then iterates through the database to find any remaining messages
+1. It then sends all the buffered messages and switches to realtime flow
+
+### Sharding
+
+There are two possible sharding strategies.
+
+- By topic index
+- By clientid hash
```
thalesmg, line 112: Maybe this could be a configuration depending on the use case?

I was thinking that retained messages are more likely to be accessed by topics, while persistent sessions are more akin to the clientids of subscribers.

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,164 @@
...
+- Load-balancer friendly.
+  Clientid is known when the client connects, so the LB can connect the client to the correct shard directly.
+
+Cons:
+- Replaying is more tricky, since the messages matching the topic filter can be found on any nodes in the cluster.
+  Replaying the messages will require starting a "replay client" process on all nodes and developing a simple protocol for coordinating the clients.
+
+### Replication
+
+TBD.
```
thalesmg, line 142: Possibly interesting references:

- https://github.com/facebook/rocksdb/wiki/Replication-Helpers
- https://github.com/pinterest/rocksplicator/
- https://medium.com/pinterest-engineering/open-sourcing-rocksplicator-a-real-time-rocksdb-data-replicator-558cd3847a9d

active/0021-rocksdb-message-persistence.md
ieQu1: It's about the same client handler. I lean towards sharding by publisher clientid (or some causality token independent from the topic), so while there are indeed multiple client handlers per shard, they can work independently as there's no causality between messages in different shards. I will update the description.

active/0021-rocksdb-message-persistence.md
ieQu1: Thanks, it was unnecessary in this case actually. I removed it.

active/0021-rocksdb-message-persistence.md
```diff
@@ -0,0 +1,335 @@
...
+
+
+## Testing Suggestions
+
+- Use [snabbkaffe](https://github.com/kafka4beam/snabbkaffe) for integration tests
+- Use [concuerror](https://github.com/parapluu/Concuerror/) to check replication and replay algorithms
+
+## Declined Alternatives
+
+### Interleaving time and topiv: topic layer hashes + time, partition by time
```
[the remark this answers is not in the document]
k32: Rocksdb doesn't recommend to have more than a hundred of column families. The partitioning will be very coarse, so lookups will be inefficient.
Epoch + offset achieves the same with more elegance.

zmstone (approved):

PR #70, opened 2022-09-21
feat: add 0021, file delivery over mqtt

id:


active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+
+* Non-reliable transmission is not supported
+* No support for one-to-many transfers
+* No distributed cache storage
+* Broker does not aware of interrupted transfer, It will be client-side implementation
+
+### Limitations
+
+* File transfer is PUSH mechanism but not support POLL, the receiving side cannot request to initiate the transfer
+decisions of the size of a slice cannot be as natural as streaming.
```
thalesmg, line 85: Should this be a separate item?  Or is it indeed part of the previous phrase in the previous line?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* Simultaneous transfer of MQTT and various types of files over a single network connection
+* Separate data transfer and file transfer channels to achieve non-blocking transmission
+  * Channel priority setting, specify the priority of different channels according to customer requirements
+  * Channels can cooperate with each other to achieve overall flow control
+  * Multiple channels can be specified for file transfer at the same time to increase file transfer parallelism
+* Efficient transmission support in weak network
+  * File transfer in weak networks such as Telematics and mobile
+  * Automatic file chunking, block-based granularity breakpoint transfer (or file-level breakpoint transfer?)
+  * Switch to TCP if QUIC is not available
+
```
thalesmg, line 42: Q: is this supposed to be for 5.0?  4.x too?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+## Design
+
+In the first phase we are going to focus on limited scope:
+
+* device-to-cloud file transfer only
+* reusing existing TCP connection
+* push only, i.e. broker cannot request file transfer, but device can start sending files any time
+* device may not know total file size when starting the transfer
+
```
thalesmg, line 57: Q: if a device uploads a file and, before it publishes to the `/fin` topic to finalize the transfer, it crashes and "forgets" that it has already uploaded something.  Would we have a periodic sweeper process that clears those orphan files / segments?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* RestAPI interface Query transfer status
+* EMQX local incomplete file cache
+* Client API should be easy to use: easy to migrate from existing applications (ftp, nfs, s3)
+* File body Encryption and decryption (with external key management)?
+
+### Out of scope
+
+* Non-reliable transmission is not supported
+* No support for one-to-many transfers
+* No distributed cache storage
```
thalesmg, line 79: This means that the device must send all the payloads to the same broker, right?  That would require correct load-balancer configuration for the session to be sticky, I guess?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+manually, list them here.
+
+## Innovation Opportunities
+
+* Integration with rule engine
+* Support for more back-end, database writes
+* Multiple backend writes to a single file
+* Support for rules processing of metadata
+* Generate modified properties of target files based on file metadata and rules of rules engine
+* Trigger event data when file upload starts/completes Enter rule chain for rule engine
```
thalesmg, line 158:
```suggestion
* Trigger event data when file upload starts/completes/enter rule chain for rule engine
```
?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+
+* 2022-09-20: @qzhuyan, @id Initial draft
+
+## Abstract
+
+TODO:
+
+## Motivation
+
+EMQX customers are asking for file transfer functionality from IoT devices to their cloud (primary use case), and from cloud to IoT devices over MQTT. Right now they are uploading files from devices via FTP or HTTPS (e.g. to S3), but it's not working well. For example, both FTP and HTTP servers usually struggle to keep with large number of simultaneous bandwith-intensive connections. Another issue is that uploading files over weak/spotty/unreliable network is not a good experience due to frequent connection reset and transfer restarts.
```
zmstone, line 13: More reasons (better to transform the reasons into bullets)
- Adding features to use cases where MQTT is already in place, if other protocols are to be used, it's usually a very different tech stack to deal with. 
  - Another set of programming SDK
  - Different types of security aspects including authentication and authorisation

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* Video and audio data from truck cars, and video data captured by inbound unloading cameras
+* Vehicle real-time logging data and messaging
+* ML logs
+
+Cloud-to-device file transfer use cases:
+
+* AI/ML models
+* Firmware upgrades
+
+With EMQX, Customers can transfer real-time IoT data (e.g. structured, as well as unstructured IoT, sensor, and industrial data), and will also support offline types of bulk data transfer (video, audio, images,  compressed log files, etc.), which, combined with EMQX's rules engine, will make it extremely easy for users to connect any IoT data to the cloud.
```
zmstone, line 29:
```suggestion
With EMQX, Customers can transfer real-time IoT data (e.g. structured, as well as unstructured IoT, sensor, and industrial data), and will also support offline types of bulk data transfer (video, audio, images,  compressed log files, etc.), which, combined with EMQX's Rule Engine and Data Integration, will make it extremely easy for users to connect any IoT data to the cloud.
```

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* No support for one-to-many transfers
+* No distributed cache storage
+* Broker does not aware of interrupted transfer, It will be client-side implementation
+
+### Limitations
+
+* File transfer is PUSH mechanism but not support POLL, the receiving side cannot request to initiate the transfer
+decisions of the size of a slice cannot be as natural as streaming.
+* Due to network bandwidth limitations, it may happen that multiple files are transferred in parallel but all fail to complete the transfer in the desired time.
+* Do not get too hung up on the length of the topic and do not take advantage of "topic alias" (MQTT 5.0)
```
zmstone, line 87: this protocol is **over** MQTT, so the use or no-use of "topic alias" should not be in this level's discussion.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* Performance may be limited by disk IO (network in speed > disk out speed)
+* Overwrite existing file SEG
```
zmstone, line 89:
```suggestion
* At least once delivery, resent (existing) file segments will be overwritten.
* ```

File: active/0021-transfer-files-over-mqtt.md [hunk 10]
```diff
@@ -0,0 +1,175 @@
...
+* Linux filename length 255 bytes. path_max 4096
+* EMQX terminates file transfers, file message MQTT forwarding (as proxy) is not required.
```
zmstone, line 91:
```suggestion
* EMQX terminates file transfers, file message MQTT forwarding (as proxy) is not supported.
```

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* QUIC in EMQX does not support HTTP/3 protocol, old client supports HTTP upload, how to smooth migration?
```
zmstone, line 92: We want to have a clear boundary/scope for this project.
Move the uncertain parts to "Discussion"
```suggestion
* When using MQTT over QUIC as transport, how to take advantage of QUIC protocol (or HTTP3) is not out of the scope of this EIP.
```

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+### Protocol design
+
+* Maximum bandwidth utilization, high signal-to-noise ratio (DATA/ MQTT HEADER)
+* Asynchronous stateless non-continuous transfer, sliceable, messy (retransmission after disconnection) (parallel transfer of multiple files)
+* Fault tolerance.
+  * Incomplete packets Disconnected.
+  * Transmission can be paused or resumed for a long time
+  * Interactive completions required???
+* Local caching of file contents
+* Utilizes only MQTT topic and payload (Payload Format Indicator?). No header changes.
```
zmstone, line 103:
```suggestion
* Utilizes only MQTT topic and payload (Payload Format Indicator?). No MQTT v5 specific features such as user properties.
```

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* TCP multi-connection support (using Same Client ID)
```
zmstone, line 104: This is impossible. One MQTT client can only have one TCP connection.
At a higher level, (e.g. a client SDK), maybe they can generate client IDs

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* User could manipulate file names, and paths. (blacklist/check file names/path, or relative path only)
```
zmstone, line 105: Suggestion: 
* First publish: `$file/sha256sum/init` to initiate the transfer, and the payload can be a json body of file metadata (whatever the user wants to put in)
* Following messages: `$file/sha256sum/0` as topic name, payload for file content.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+* Symmetric transfer (client -> broker)
+* PUBACK (QoS) guarantees transmission reliability (considering that the client does not receive PUBACK and resends it after disconnection)?
```
zmstone, line 107: * Only allow `QoS=1`, PUBACK with error code can be used as failure indication.

Retransmission is already covered above.

active/0021-transfer-files-over-mqtt.md
zmstone: > Suggestion:
> 
> * First publish: `$file/sha256sum/init` to initiate the transfer, and the payload can be a json body of file metadata (whatever the user wants to put in)
> * Following messages: `$file/sha256sum/<byte-offset>` as topic name, payload for file content.

We will actually need a protocol for this init message in order to communicate the total size of the file etc.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,175 @@
...
+## Requirements
+
+* Simultaneous transfer of MQTT and various types of files over a single network connection
+* Separate data transfer and file transfer channels to achieve non-blocking transmission
+  * Channel priority setting, specify the priority of different channels according to customer requirements
+  * Channels can cooperate with each other to achieve overall flow control
+  * Multiple channels can be specified for file transfer at the same time to increase file transfer parallelism
+* Efficient transmission support in weak network
+  * File transfer in weak networks such as Telematics and mobile
+  * Automatic file chunking, block-based granularity breakpoint transfer (or file-level breakpoint transfer?)
```
jinfahua, line 40: Can the block size configured by user? For example, 1024 is the default block size, and user can set to 2048 etc.

active/0021-transfer-files-over-mqtt.md
id: Impossible because we can't/won't implement it, or impossible because it's a fundamental limitations of the protocol?

active/0021-transfer-files-over-mqtt.md
id: As far as I understand there are use cases when client would not know file size in advance, like rolling log file.

active/0021-transfer-files-over-mqtt.md
id: I don't see why not.

active/0021-transfer-files-over-mqtt.md
id: How would broker deal with file metadata? Is it going to be a separate .json with the same name?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+* Upload collected ML logs
+
+Known cases of cloud-to-device file transfer:
+
+* Upload AI/ML models
+* Firmware upgrades
+
+## Terminology
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://www.rfc-editor.org/bcp/bcp14) [[RFC2119](https://www.rfc-editor.org/rfc/rfc2119)] [[RFC8174](https://www.rfc-editor.org/rfc/rfc8174)] when, and only when, they appear in all capitals, as shown here.
```
id, line 35: Not using these everywhere yet, and when I do, not always correctly. Will fix.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+As an example of existing implementation we can look at AWS IoT Core [which provides functionality](https://docs.aws.amazon.com/iot/latest/developerguide/mqtt-based-file-delivery.html) to [deliver files to IoT devices](https://docs.aws.amazon.com/iot/latest/developerguide/mqtt-based-file-delivery-in-devices.html):
+
+![](0021-assets/aws-mqtt-file-delivery.png)
+
+## Design
+
+### Overview
+
+* Files are split in segments of equal length with the exception of the last segment, it's length is > 0 and  <= segment length
+* Client generates UUID for each file being transferred and use it as file Id in Topic Name
```
id, line 71: Not sure it's the right approach.
My thinking was as follows:
* we need some kind of unique file identifier anyway
* file path + file name could be fairly large, sending them as part of PUBLISH message with each segment could be greedy
* UUID can be stored in the session and broker can use it to get access to the rest of file metadata

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+
+As an example of existing implementation we can look at AWS IoT Core [which provides functionality](https://docs.aws.amazon.com/iot/latest/developerguide/mqtt-based-file-delivery.html) to [deliver files to IoT devices](https://docs.aws.amazon.com/iot/latest/developerguide/mqtt-based-file-delivery-in-devices.html):
+
+![](0021-assets/aws-mqtt-file-delivery.png)
+
+## Design
+
+### Overview
+
+* Files are split in segments of equal length with the exception of the last segment, it's length is > 0 and  <= segment length
```
id, line 70: equal length is actually not critical, but if we don't do that, then we need to send segment offset together with payload, and the built-in MQTT mechanism of connecting PUBACK to corresponding PUBLISH messages via Packet Identifier might not always work, since Packet Identifier field is only 2 bytes.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+* Client generates UUID for each file being transferred and use it as file Id in Topic Name
+* Client calculates sha256 checksum of the segment it's about to send and sends it as part of Topic Name
+* Client uses $file Topic Filter to transfer files
+* Clients cannot subscribe to $file topics
```
id, line 74: this implicitly tells that we do not support 'proxy' mode for file transfers

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+  * Payload Format Indicator=0x01
+  * `{fileId}` is corresponding file UUID
+  * Payload is a JSON document with the following properties:
+    * filepath
+      * a character string composed with characters from this set [a-zA-Z0-9_-./]
+      * length in bytes MUST be <= 4096 (PATH_MAX=4096 on most common Linux filesystems)
+    * filename
+      * a character string composed with characters from this set [a-zA-Z0-9_-./]
+      * length in bytes MUST be < 256 (NAME_MAX=255 in Linux)
+    * size
```
id, line 99: this can be optional

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+| OMIT        |                               | Same as 0x00                                        |
+| 0x00        | Success                       | File segment has been successfully persisted        |
+| 0x10        | No matching subscribers       | Server asks Client to retransmit all segments       |
+| 0x80        | Unspecified error             | Server asks Client to retransmit a specific segment. Segment sequential number is indicated by Packet Identifier field in the PUBACK Variable Header |
+| 0x83        | Implementation specific error | Server asks Client to cancel the transfer           |
+| 0x97        | Quota exceeded                | Server asks Client to pause the transfer            |
+
+#### 0x83, "Implementation specific error", "Cancel Transfer"
+
+Client can retry the transfer after 1 hour.
```
id, line 140: arbitrary number, open for suggestions

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+
+#### 0x97, "Quota exceeded", "Pause Transfer"
+
+Upon receiving PUBACK message with this reason code, Client is expected to delay sending next packet for 10 seconds. If next PUBACK is also 0x97, Client delays for 20 seconds. Client continues to double the delay length until it reaches 80 seconds.
```
id, line 144: arbitrary numbers again, suggestions are welcome

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+* Topic Filter
+* MQTT Control Packet
+
+*At least once*: a message can be delivered many times, but cannot be lost
+
+## Requirements
+
+* The protocol MUST use only PUBLISH type of MQTT Control Packet
+* The protocol MUST support transfer of file segments
+* Server MUST be able to verify integrity of each file segment
```
zmstone, line 51: Is there a way to verify the integrity after putting all segments together?
Everything ends, right? even if it's a stream.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+* Client MAY know total file size when initiating the transfer
+* Client MAY abort file transfer
+* Server MAY ask the client to pause file transfer
+* Server MAY ask the client to abort file transfer
+* The protocol MUST NOT require changes in client code
```
zmstone, line 56: I find it odd to put this here.  Maybe one requirement is that this portocol is on top of MQTT (which implies this is not imposing changes any existing MQTT client SDK). 
However I still find it odd to put implementation options in a protocol specification.
```suggestion

```

active/0021-transfer-files-over-mqtt.md
id: final sha256 as part of the `fin` message? wdyt?

active/0021-transfer-files-over-mqtt.md
zmstone: 1. Let's not paint ourselve in this the corner or fixed size packet.
2. Let's avoid using Packet ID for other purposes than coorelating sent/ack (which is at MQTT layer, and we are on hihger layer). 
  * When using an SDK, packet ID is very much likely abstracted away.
  * For app layer controlled re-transmision, the same segment can be sent with two different packet IDs
  * The packet ID goes back to 0 after overflow
  * The packet ID can be set back to zero after idling in some implementations

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,206 @@
...
+
+Initialize the file transfer. Server is expected to store metadata from the payload in the session along with `{fileId}` as a reference for the rest of file metadata.
+
+  * Qos=1
+  * DUP=0 for the initial segment transfer, 1 when retransmitting
+  * Payload Format Indicator=0x01
+  * `{fileId}` is corresponding file UUID
+  * Payload is a JSON document with the following properties:
+    * filepath
+      * a character string composed with characters from this set [a-zA-Z0-9_-./]
```
savonarola, line 93: What is the reason for the character set to be limited?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,206 @@
...
+      * total file size in bytes
+
+#### `$file/{fileId}/{offset}/{sha256sum}` message
+
+One such message for each file segment.
+
+  * Qos=1
+  * DUP=0 for the initial segment transfer, 1 when retransmitting
+  * Payload Format Indicator=0x00
+  * Packet Identifier=sequential file segment number
```
savonarola, line 108: Does this comply with the protocol? If we send several files, packet identifier will duplicate?

savonarola:
Do we consider Client <-> Client file transfer? Or, maybe, Broker -> Client file transfer?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,205 @@
...
+* Client uses $file Topic Filter to transfer files
+* Clients cannot subscribe to $file topics
+* Segment length can be calculated on the server side by subtracting the length of the [Variable Header](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901025) from the [Remaining Length](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901105) field that is in the [Fixed Header](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901021)
+* Data is transferred in PUBLISH packets in the following order:
+  1. $file/{fileId}/init
+  2. $file/{fileId}/{sha256sum}
+  3. $file/{fileId}/{sha256sum}
+  4. ...
+  3. $file/{fileId}/[fin|abort]
+* Client can send up to N (configured) PUBLISH packets before blocking for PUBACK
```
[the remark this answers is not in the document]
id: TODO: remove this, MQTT clients have built-in in-flight window config

active/0021-transfer-files-over-mqtt.md
id: TODO: remove this requirement

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,206 @@
...
+  * Qos=1
+  * DUP=0 for the initial segment transfer, 1 when retransmitting
+  * Payload Format Indicator=0x01
+  * `{fileId}` is corresponding file UUID
+  * Payload is a JSON document with the following properties:
+    * filepath
+      * a character string composed with characters from this set [a-zA-Z0-9_-./]
+      * length in bytes MUST be <= 4096 (PATH_MAX=4096 on most common Linux filesystems)
+    * filename
+      * a character string composed with characters from this set [a-zA-Z0-9_-./]
```
id, line 96: TODO: remove this requirement

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,206 @@
...
+      * length in bytes MUST be < 256 (NAME_MAX=255 in Linux)
```
id, line 97: TODO: remove this requirement

active/0021-transfer-files-over-mqtt.md
id: TODO: remove this requirement

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,206 @@
...
+
+One such message for each file segment.
+
+  * Qos=1
+  * DUP=0 for the initial segment transfer, 1 when retransmitting
+  * Payload Format Indicator=0x00
+  * Packet Identifier=sequential file segment number
+  * Payload is file segment bytes
+  * `{offset}` is byte offset of the given segment
+  * `{sha256sum}` is sha256 checksum of file segment bytes
```
id, line 111: * can we have offset and sha256 as part of the payload?
** upside: static and short topic names
** upside: clients can use topic alias if they support it
** downside: clients have to assemble payload, we need a serialization format

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,206 @@
...
+  * Qos=1
+  * no payload
+
+### PUBACK Reason codes
+
+| Reason code | MQTT Name                     | Meaning in file transfer context                    |
+|-------------|-------------------------------|-----------------------------------------------------|
+| OMIT        |                               | Same as 0x00                                        |
+| 0x00        | Success                       | File segment has been successfully persisted        |
+| 0x10        | No matching subscribers       | Server asks Client to retransmit all segments       |
```
id, line 134: * could be used when e.g. file retention kicks in while client is still trying to send some later segments

id (commented):
add diagram when broker received the same file segment

id:
Checksum for each segment and the file
* do we need it if we use TLS?
* cheaper checksum algorithms?

active/0021-transfer-files-over-mqtt.md
id: server will have some max retention and will be killing unfinished segments

TODO: add this to the spec

active/0021-transfer-files-over-mqtt.md
id: Brokers will be storing segments in local cache and gather segments when file transfer is finalized based on information in the session.
This is implementation specific, we probably don't want to add this into the spec.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,233 @@
...
+  * Payload Format Indicator=0x01
+  * `{fileId}` is corresponding file UUID
+  * Payload is a JSON document
+
+##### `init` payload JSON Schema
+
+```json
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://emqx.io/file-over-mqtt.init.schema.json",
```
zmstone, line 102: we should put this to a git repo, and put a git tag on it, then reference it from here?

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,233 @@
...
+  "title": "init",
+  "description": "Schema for payload of the init message",
+  "type": "object",
+  "properties": {
+    "name": {
+      "description": "File name",
```
zmstone, line 108:
```suggestion
      "description": "A descriptive name for display",
```

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,233 @@
...
+    "size": {
+      "description": "Total file size in bytes, optional.",
+      "type": "integer"
+    },
+    "expire_at": {
+      "description": "File expiration time as UNIX Epoch.
+                      After this point in time broker can delete the file.",
+      "type": "integer"
+    },
+    "segment_ttl": {
```
zmstone, line 120:
```suggestion
    "segments_ttl": {
```

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,233 @@
...
+      "description": "Time-to-leave (in seconds) for segments of incomplete file transfer.
+                      Broker will garbage collect segments of unfinished transfers when
+                      this time expires. If not set, broker will use its own default value.",
```
zmstone, line 123:
```suggestion
      "description": "Time-to-live (in seconds) for segments of incomplete file transfer.
                      The deadline of for broker to collect all segments is calculated from InitTime + TTL,
                      where InitTime is the time when broker receives the init message. 
                      Broker will garbage collect segments of unfinished transfers when
                      this time expires. If not set, broker will use its own default value.",
```

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,233 @@
...
+    "user_data": {
+      "description": "User-defined metadata, optional.",
+      "type": "object"
+    }
+  },
+  "required": [ "name", "expire_at" ]
+}
+```
+
+* Broker SHOULD have default setting for `segment_ttl`
```
zmstone, line 135:
```suggestion
* Broker SHOULD have default setting for `segments_ttl`
```

active/0021-transfer-files-over-mqtt.md
id: I suck with naming. If you have a better idea what to call https://github.com/emqx/mqtt-file-transfer-schema, let me know.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,199 @@
...
+![](0021-assets/aws-mqtt-file-delivery.png)
+
+## Design
+
+### Overview
+
+* Files are split in segments, segments can be of arbitrary length
+* Client MUST generate UUID according to [RFC 4122](https://www.rfc-editor.org/rfc/rfc4122) for each file being transferred and use it as file Id in Topic Name
+* Broker MUST validate that provided UUID conforms to [RFC 4122](https://www.rfc-editor.org/rfc/rfc4122)
+* Client calculates sha256 checksum of the segment it's about to send and sends it as part of Topic Name
```
zmstone, line 75: We should make this optional.
yesterday we discussed that doing sha256 checksum is quite an expensive thing.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,199 @@
...
+  * Payload Format Indicator=0x01
+  * `{fileId}` is corresponding file UUID
+  * Payload is a JSON document
+
+##### `init` payload JSON Schema
+
+Available [here](https://github.com/emqx/mqtt-file-transfer-schema/blob/v1.0.0/init.json).
+
+* Broker SHOULD have default setting for `segments_ttl`
+* Broker MAY delete segments of unfinished file transfers when their TTL has expired
```
zmstone, line 102: We should probably add an optional field to hold the assembled sha256 checksum.
And make the per-segment checksum optional too.

active/0021-transfer-files-over-mqtt.md
```diff
@@ -0,0 +1,199 @@
...
+  3. $file/{fileId}/{offset}/{sha256sum}
+  4. ...
+  3. $file/{fileId}/[fin/{sha256sum} | abort]
+
+#### `$file/{fileId}/init` message
+
+Initialize the file transfer. Server is expected to store metadata from the payload in the session along with `{fileId}` as a reference for the rest of file metadata.
+
+  * Qos=1
+  * DUP=0 for the initial segment transfer, 1 when retransmitting
```
zmstone, line 92: Not sure how useful is the DUP flag.
Re-sedning a init message for an existing ID should result in a purge of old file (or segments if it was not complete).

zmstone:
Speaking of re-transmition.
If the client decies to re-send a segment which overlaps with other already sent segments.
e.g. in case of streaming,
after `{fileid}/1000/{checksum}` is sent with `1000` bytes payload, but the connection was broken befre PUBACK.
some more bytes accumulated, then the client tries to send `{fileid}/1000/{checksum}` with more than `1000` bytes.

another case:
1.  send `offset=1000, payload=1000`
2. send `offset=2000, payload=1000`
3. send `offset=1000, payload=2000`
i.e. 3 is an attempt to overwrite 1 and 2.

id:
renamed commit messages to follow the same format

ieQu1:
From my experience with working SFTP protocol:

1) It's not advised to send chunks synchronously, since network latency will affect the throughput. From the message sequence diagrams I see that this problem has been addressed and the client doesn't wait for PUBACKs before sending new chunks, so protocol follows "streaming" pattern. Good.
2) At the same time, streaming requires backpressure to avoid unnecessary retransmissions when the broker's cache fills up. From the quick look at the EIP I didn't find any explanation as to how backpressure will be applied. Could you clarify this part?

qzhuyan:
@ieQu1 
for 
> At the same time, streaming requires backpressure to avoid unnecessary retransmissions when the broker's cache fills up. From the quick look at the EIP I didn't find any explanation as to how backpressure will be applied. Could you clarify this part?

use error code ```| 0x97        | Quota exceeded                | Server asks Client to pause the transfer            |```

qzhuyan (approved):

ieQu1 (approved):

PR #71, opened 2022-12-07
feat(0022): forward compatibility check for install upgrade script (5.0)

thalesmg:
https://emqx.atlassian.net/browse/EMQX-6598


qzhuyan:
what about the downgrade, shall it perfer to use the script from old version, I guess not?

thalesmg:
> what about the downgrade, shall it perfer to use the script from old version, I guess not?

Good question.  Since the newer script supposedly contains fixes that the older one doesn't contain, I think it'd be better to use the newer one for downgrading.

zmstone (approved):

Issue #59, opened 2021-12-09
local session resume


zmstone:
Today, when a client re-connects. A new connection process is spawned (= a new session).
the new session try to take over the old, the old will unsubscribe itself, and the new will re-subscribe (in the trie table).
This is done even if the old session is found in the local node.

This issue is opened to
1. Investigate if it is feasible to handover the socket to the old process.
2. If 1 is feasible, create a proposal to follow up the changes.

PR #76, opened 2023-06-27
feat: proposal for hybrid session takeover and resumption in emqx 5

emqplus:
fixes: https://github.com/emqx/eip/issues/59


zmstone:
Great proposal.

Not sure if QUIC socket owner handover is supported as TCP/TLS socket, if not, we'll need to treat them differently.
@qzhuyan ?

zmstone:
In the current implementation, if we only look at the takeover initialization phase, it consists of 2 steps

1. Sends a nice step-down request to the old
2. If step 1 timeout, force shutdown (by kill the old)

We'll need to consider the scenario of timeouts during local resume (socket handover) too.
Some exception flows to cover:

* If the old process receives a new socket while the old socket still active (e.g. no tcp_closed was received), how to proceed?
  Maybe close the old and accept the new, then we'll need to take care of the state transitions as if the old socket is closed in a normal way.
* If the old process is not responding to socket handover, how to proceed?
  When timeout happens, good chance the process got stuck somewhere (e.g. calling out to external database for authentication), and the old socket close or error event is maybe in its mailbox already.

zmstone (approved):

PR #79, opened 2023-09-28
feat(0023): OpenTelemetry trace integration - implemented

SergeTupchiy:


active/0024-opentelemetry-trace-integration.md
```diff
@@ -0,0 +1,211 @@
...
+
+- internal tracing API (as of now, implemented in `emqx_otel_trace` module) is not decoupled from the rest of the code base: each traceable action (span) is traced by a specific function and all these functions are quite specific. For example, they may extract/propagate the context differently and/or rely on the previous (parent) span. For now, it doesn’t seem feasible to create a generic trace wrapper that can trace an arbitrary function.
+- current implementation modified `mqtt_packet` record by adding an `internal_extra` field to trace some events occurring after `emqx_message` is converted to `mqtt_packet`:
+
+  ```erlang
+  -record(mqtt_packet, {
+    ...
+    %% Experimental, added to carry Opentelemetry context after message is converted to packet,
+    %% must always be ignored when serialized, as it is not a part of MQTT packet
+    internal_extra = [] :: list()
```
zmstone, line 101: can we keep tracing in user properties ?
i.e. `#mqtt_packet.variable`

active/0024-opentelemetry-trace-integration.md
```diff
@@ -0,0 +1,211 @@
...
+  }).
+  ```
+
+![An actual EMQX trace example as exported by POC implementation](0024-assets/trace-export-example.png)
+
+### External trace context propagation
+
+If EMQX receives trace context in a published message, e.g., `traceparent`/`tracestate` User-property for MQTT v5.0, it must be sent unaltered when forwarding the Application Message to a Client to conform with [MQTT specification 3.3.2.3.7](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901116).
```
zmstone, line 109: > The Server MUST send all User Properties unaltered

But the server can probably add additional (new) properties -- if it helps.

SergeTupchiy: Yes, we probably can add trace context created by EMQX to User properties, so the consumer can continue tracing if needed, that's should be fine.

active/0024-opentelemetry-trace-integration.md
SergeTupchiy: Yes, but if so, I think we would mix our custom internal data (that shouldn't be serialized) together with MQTT specific properties :thinking:
Another idea I had is to wrap `mqtt_packet` record with some additional structure. Can be as simple as: `{#mqtt_packet{}, Extra}` but it might be error prone.

SergeTupchiy: Changed,  `internal_extra` field in `properties` map will be excluded when serialized to frame:
```
serialize_property(_, Disabled) when Disabled =:= disabled; Disabled =:= undefined ->
    <<>>;
serialize_property(internal_extra, _) ->
    <<>>;
serialize_property('Payload-Format-Indicator', Val) ->
    <<16#01, Val>>;
...
```

active/0024-opentelemetry-trace-integration.md
```diff
@@ -0,0 +1,211 @@
...
+
+-spec trace_dispatch(Topic, Msg, fun((Topic, Msg) -> Res)) -> Res when
+    Topic :: emqx_types:topic(),
+    Msg :: emqx_types:message(),
+    Res :: emqx_types:deliver_result().
+trace_dispatch(Topic, Msg, DispatchFun) ->
+   ...
+
+put_ctx_to_msg(OtelCtx, Msg = #message{extra = Extra}) ->
+    Msg#message{extra = lists:keystore(?EMQX_OTEL_CTX, 1, Extra, {?EMQX_OTEL_CTX, OtelCtx})}.
```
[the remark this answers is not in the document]
SergeTupchiy: Done

active/0024-opentelemetry-trace-integration.md
```diff
@@ -0,0 +1,214 @@
...
+
+  - Route message (traced by a node that received a published message)
+
+    - Dispatch message (traced by all nodes that have matched subscribers)
+
+      - Deliver/send message to a subscriber (traced by all nodes that have matched subscribers)
+
+NOTE: the above list may be extended/changed, for now it follows the product requirements.
+
+Any other processing steps/events like client connection, authentication, subscription are currently not considered for OpenTelemetry tracing (**TBD**) due to the following reasons:
```
savonarola, line 48: Have we decided on:
* tracing authorization (this seems to be a part of publish flow)
* tracing puback, pubrel, pubcomp within the same trace id as publishes
* the scope/kind of events we are going to add to the traces?

SergeTupchiy: Not yet, but the latest result of the discussion with @zmstone is to start with a minimum number of trace spans, e.g.: 
- process a published message (from the point when the packet is received and decoded till it is dispatched locally and forwarded to other nodes)
- publish the message to a subscriber (one span per each connection on every node, starting from receiving the message by a connection process till sending it to a port).

active/0024-opentelemetry-trace-integration.md
thalesmg: Since original trace context key-pairs are to be forwarded unaltered, it sounds like that the new keys that EMQX adds would also need some kind of incremental ID attached to them, in case a message is forwarded through a chain of brokers (in different clusters)?  

e.g.: broker A receives a message (with trace context), adds its own context, forwards it to broker B, which forwards to C, etc.

active/0024-opentelemetry-trace-integration.md
```diff
@@ -0,0 +1,214 @@
...
+
+This also perfectly follows [OpenTelemetry semantics for messaging systems](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md#context-propagation):
+
+> Messaging systems themselves may trace messages as the messages travels from producers to consumers. Such tracing would cover the transport layer but would not help in correlating producers with consumers. To be able to directly correlate producers with consumers, another context that is propagated with the message is required.
+>
+> A message creation context allows correlating producers with consumers of a message and model the dependencies between them, regardless of the underlying messaging transport mechanism and its instrumentation.
+>
+> The message creation context is created by the producer and should be propagated to the consumer(s).
+>
+> A producer SHOULD attach a message creation context to each message. If possible, the message creation context SHOULD be attached in such a way that it cannot be changed by intermediaries.
```
thalesmg, line 122: This got me thinking: would messages produced as a result of rule engine processing be considered part of the same trace parent, or would be considered an entirely new trace root?

The original message might be still forwarded and processed somewhere else, but it sounds like those derived messages would initiate fresh contexts.  :thinking:

active/0024-opentelemetry-trace-integration.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Sampling
+
+OpenTelemetry sampling is described in great depth in the [official documentation](https://opentelemetry.io/docs/concepts/sampling).
+
+Erlang opentelemetry lib implements only head sampling. Head sampling implies that a sampling decision is made as early as possible, e.g., by following a configured percentage of traces to sample (100% by default). A decision to sample or drop a span or trace is not made by inspecting the trace as a whole.
+
+Sampling rate option should be added to EMQX configuration.
+
+Tail sampling that makes a sampling decision after all the spans are done would need to be implemented by extending opentelemetry lib (**TBD**).
```
thalesmg, line 154: The examples of tail sampling below sound more like they are a form of filtering rather than sampling.  Perhaps those are good candidates for filter criteria, since they are not quite random?

active/0024-opentelemetry-trace-integration.md
SergeTupchiy: In general, it's enough to propagate an original trace context unchanged through a chain of brokers (intermediaries). 
All the traces emitted by each broker will still be correlated, since they will use the same trace ID.
There will be no direct parent-child relationship between brokers in the chain, but it still can be easily tracked down by timestamps, e.g. the second broker can't emit its traces earlier than the first broker.

We can add some additional specific trace headers to keep the links between brokers in the chain, but: 
- it will be a custom extension, requiring all the intermediary brokers to implement it
- it will probably not work out of the box in the standard OpenTelemetry backends
Alternatively, we can overwrite the original trace context: it will still keep the original trace ID but use the latest span ID, so that the second broker trace is linked directly to the last traced span of the first broker. 
However, it would:
- violate MQTT spec (User property would overwritten) 
- not follow [OpenTelemetry semantic conventions for messaging systems](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md#intermediary): 
    > A producer SHOULD attach a message creation context to each message. If possible, the message creation context SHOULD be attached in such a way that it cannot be changed by intermediaries.

@thalesmg , please correct me if I have not understood your idea exactly :thinking: :slightly_smiling_face:

active/0024-opentelemetry-trace-integration.md
SergeTupchiy: Head/tail sampling is OpenTelemetry terminology.
Tail sampling is essentially stateful: it depends on the given trace events, conditions occurred some time after the trace started. 
Filtering is EMQX specific, inspired by EMQX log tracing, it must stateless and quite simple (if need it at all). 
I agree that both concepts are very similar, especially when sampling means non-probability based approaches.
Probably, they should be somehow unified but we also need to define which ones need to be implemented (at least in the first iteration).

active/0024-opentelemetry-trace-integration.md
SergeTupchiy: This a good question, I think both approaches are possible.

With the latest changes agreed with @zmstone (EIP not updated yet), we are going to trace only two spans, the first one 
starting at receiving PUBLISH packet: https://github.com/emqx/emqx/pull/11696/files#diff-27c0844d0e2116cd96569c34369a4b485d68529f0d21cb075c19f8bc5a59f477R402
This first traced span will also extract `traceparent` from `User-Properties`, the next trace span (sending a published message to a subscriber) will depend on the already extracted context (it won't try to extract `traceparent` from 'User-Properties' again).

That means that a republished message will never be traced, since it starts from calling `emqx_broker` directly, so it won't trigger trace context extraction/creation, which is probably fine for now, as we plan to keep it simple in the first iteration.

ieQu1 (approved):

PR #80, opened 2023-10-10
feat(ft): add async mode for file transfer

savonarola:
In async mode, file transfer command execution result is published as an MQTT message with JSON payload to a special topic.


implemented/0021-transfer-files-over-mqtt.md
```diff
@@ -101,25 +101,84 @@ Data is transferred in PUBLISH packets in the following order:
...
+
+### Sync mode vs Async mode
+
+File transfer individual commands may be handled in two modes: syncronous and asyncronous. The mode is chosen by the client, by sending commands either to `$file/...` (sync mode) or `$file-async/...` (async mode) topics.
+
+Client is free to mix arbitrarily sync and async commands in the same file transfer session.
+
+#### Syncronous mode
+
+In syncronous mode, the successful/unsuccessfull status of individual operations is communicated to the client via Reason Code field of MQTT `PUBACK` messages. See [Reason codes](#reason-codes) for details.
```
keynslug, line 131:
```suggestion
In syncronous mode, the successful/unsuccessful status of individual operations is communicated to the client via Reason Code field of MQTT `PUBACK` messages. See [Reason codes](#reason-codes) for details.
```

implemented/0021-transfer-files-over-mqtt.md
```diff
@@ -101,25 +101,84 @@ Data is transferred in PUBLISH packets in the following order:
...
+* Broker responds with `PUBACK`. Nonzero Reason Code indicates immediate failure.
+* Zero Reason Code indicates that the command has been accepted for processing. The client is expected to wait for the actual result of the command via `$file-response/{clientId}` topic.
+
+![](./0021-assets/flow-async-1.png)
+
+The result of the command is a JSON document with the following fields:
+```json
+{
+  "vsn": "0.1",
+  "topic": "$file/[COMMAND]",
```
keynslug, line 155: Shouldn't that be?
```suggestion
  "topic": "$file-async/[COMMAND]",
```

implemented/0021-transfer-files-over-mqtt.md
```diff
@@ -81,15 +81,15 @@ As an example of existing implementation we can look at AWS IoT Core [which prov
...
+* Client MUST use Topic starting with `$file/`(`$file-async/`) to transfer files.
+* Broker MUST NOT let clients subscribe to Topics starting with `$file/`(`$file-async/`) topics
 * Segment length can be calculated on the server side by subtracting the length of the [Variable Header](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901025) from the [Remaining Length](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901105) field that is in the [Fixed Header](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901021)
 
 ### Protocol flow
 
 Data is transferred in PUBLISH packets in the following order:
 
-1. `$file/{fileId}/init`
+1. `$file[-async]/{fileId}/init`
```
keynslug, line 92: Q: A pity that we'd need another system topic prefix for that. What do you think about moving those kind of things at the end of the topic, in a http-query-string fashion? I.e. `$/file/{fileId}/init/?mode=async`, or `$/file/{fileId}/init//mode=async` (somewhat easier to deconstruct), or something similar. This would also provide a kind of forward compatibility. Apparent downside is that commands with optional checksums at the end would be slightly more complex to parse / handle.

implemented/0021-transfer-files-over-mqtt.md
```diff
@@ -101,25 +101,84 @@ Data is transferred in PUBLISH packets in the following order:
...
+  "reason_code": 0,
+  "reason_description": "success"
+}
+```
+
+* `vsn` — response document format version;
+* `topic` — the topic of the command that the response is for, e.g. `$file-async/somefileid/init`;
+* `packet_id` — the MQTT packet id of the command packet (`PUBLISH`) that the response is for;
+* `reason_code` — the result code of the command execution. See [Reason Codes](#reason-codes) for details;
+* `reason_description` — the human-readable description of the Reason Code.
```
keynslug, line 166: Micronit: would likely look better as a table.

keynslug (commented):
👍🏼

implemented/0021-transfer-files-over-mqtt.md
savonarola: I considered some variants with mode/etc encoding as key-value pairs. However, it finally seemed to me that encoding something complex into topics somewhat contradicts the MQTT idea of topics, which are built from separators and opaque tokens.
With `/mode=.../` we e.g. cannot match against mode with standard topic matchers (`+`, `#`), cannot impose ACL, etc.

implemented/0021-transfer-files-over-mqtt.md
savonarola: ✅

implemented/0021-transfer-files-over-mqtt.md
```diff
@@ -101,25 +101,86 @@ Data is transferred in PUBLISH packets in the following order:
...
+![](./0021-assets/flow-sync-1.png)
+
+Caveats:
+
+* Some operations (`fin` command) may take considerable time to complete. So if a client wants to utilize the session while waiting for the result, it should either use async mode or implement some kind of asynchronous logic itself, that is to deal with several unacked PUBLISH messages in parallel, like this:
+
+![](./0021-assets/flow-sync-2.png)
+* MQTTv3 clients do not support reason codes, so async mode is the preferred option for them.
+
+#### Asyncronous mode
```
zmstone, line 142: Asynchronous

implemented/0021-transfer-files-over-mqtt.md
keynslug: Fair point. What do you think about a kind of middle ground where we define _mode_ as additional "positional argument", so to say? This can still provide us with forward compat, admittedly somewhat limited. E.g. `$file/abc-def/fin/4242//async` for a FIN command, where checksum is not specified. We could assume sync mode by default and say that `$file/abc-def/fin/4242//` is equivalent to `$file/abc-def/fin/4242` as a backward compat measure.

id (approved):
👍

PR #81, opened 2023-10-31
0024 and 0025: message validation and transform

zmstone:


active/0024--message-validation-and-transform.md
```diff
@@ -0,0 +1,162 @@
...
+- Message filtering
+
+This data processing step should happen after authorization and before rules engine.
+Probably can make use of the 'message.publish' with a hook priority which
+positions the callback after authorization, but before data bridges.
+
+### High level design
+
+* Introduce 2 different data processing types: `validation` and `tranform`.
+* Make use of the rules engine SQL, but without the `FROM` caluse as it's always the MQTT message being th input.
```
ieQu1, line 57:
```suggestion
* Make use of the rules engine SQL, but without the `FROM` clause as it's always the MQTT message being the input.
```

active/0024--message-validation-and-transform.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+### Message Validation
+
+```
+{
+    id = id1
+    topic = 't/1'
+    type = validation
+    action_if_invalid = disconnect # or 'drop'
```
id, line 81: If we use `message filtering` as common denominator for both validation and filtering, we could have instead `action_if_passed` and `action_if_failed`.

active/0024--message-validation-and-transform.md
```diff
@@ -0,0 +1,162 @@
...
+        description = "drop message if it is not compabitble with my avro schema"
+        topic = 't/1'
+        type = validation
+        action_if_invalid = disconnect
+        sql = "SELECT schema_decode(my_avro_schema, payload) as decoded WHERE decoded != undefined"
+    },
+    {
+        description = "increment payload.value"
+        topic = 't/2'
+        type = tranform
```
id, line 140:
```suggestion
        type = transform
```

active/0024--message-validation-and-transform.md
```diff
@@ -0,0 +1,162 @@
...
+]
+```
+
+## Backwards Compatibility
+
+New feature.
+
+## Document Changes
+
+Docs should be created to guid users to use the new feature.
```
id, line 152:
```suggestion
Docs should be created to guide users to use the new feature.
```

active/0024--message-validation-and-transform.md
```diff
@@ -0,0 +1,162 @@
...
+        topic = 't/1'
+        type = validation
+        action_if_invalid = disconnect
+        sql = "SELECT schema_decode(my_avro_schema, payload) as decoded WHERE decoded != undefined"
+    },
+    {
+        description = "increment payload.value"
+        topic = 't/2'
+        type = tranform
+        sql = "SELECT * as new, payload.value + 1 as new.payload.value"
```
keynslug, line 141: My 2 cents. Given that Rule SQL semantics in general provide a DSL to describe how to transform 1 message into 0 or 1 results (or even ≥0 with `FOREACH`), it's not obvious why this type distinction is needed. If running some SQL over a message produces 0 results (or fails to compute) it filters it out in some sense, otherwise transforms. Besides, with this type distinction the SQL syntax in different contexts might need to be constrained (e.g. disallow `FOREACH` in filters, disallow `WHERE` in transforms), which can be unintuitive for the user.

terry-xiaoyu:
I believe that introducing an entirely new message processing system outside of the rule engine would increase the overall system's complexity and the learning curve for users. I suggest expanding the capabilities of the rule engine by adding a few configuration options to allow the rule engine to terminate message flow and alter the original message content.

In my personal experience, most users do not need to terminate or modify messages in transit. However, if a user has such a requirement, they typically want all rules to have the capability to modify message flow, rather than requiring a few specific rules to refrain from altering message flow. Therefore, we can introduce a global configuration option in the rule engine's settings, something like `rule_engine.message_flow_intrusive = boolean()`, with a default value of `false`, indicating that it does not affect message flow.

With the concept of intrusive message flow, we must define the sequence of flow. We don't need to handle hooks because we've already hardcoded the order of hooks. However, the order of rules still needs to be determined. Therefore, we can add a global configuration in the rule engine: `rule_engine.message_flow_order = ["rule_name_001", "rule_name_002"]`.

For each rule, we can add a configuration, for example: `rule.rule_name_001.message_flow_action_if_matched = "stop_message_flow" | "disconnect_connection" | "none"`. This will terminate message flow or disconnect the connection after executing all actions, where "none" indicates taking no action, and the message will continue to flow to the next rule and then to other plugins if "rule_name_001" is not matched.

active/0024--message-validation-and-transform.md
```diff
@@ -0,0 +1,162 @@
...
+
+However, currently rules are only used in data bridges, but it does not interfere
+MQTT message pub/sub.
+For example, if there is subscriber subscribing to `t/#` it would still receive the avro encoded message.
+
+If we also want to stop the message being published, then there would be a need for a message republish action,
+and the publisher and subscriber will likely be forced to use different topics.
+e.g. publisher publishes to `t1/...`, republish action send it to `t2/...` for subscribers to receive.
+
+This can work, but not easy to use.
```
keynslug, line 35: Random thought, you probably have already pondered it. We could introduce something like _rewrite_ action in the Rule Engine and grant it an ability to erase the original message and replace it with one or more other messages, or even terminate the connection if configured accordingly.

active/0024--message-validation-and-transform.md
zmstone: do you mean there is no need for 'transform' vs. 'validation' ?
  --- it's mostly for config clarity, e.g. when it's `validation`, there there must be `action_if_invalid`.

or do you mean there is no need for 'transform' vs. 'filter'? 
  --- I don't think 'filter' type is proposed in this doc. "message filter" is just one of the input requirements, it's not mapped to a process type.

active/0024--message-validation-and-transform.md
zmstone: Can maybe put this whole new thing to `rule_engine.validation_and_transform`.
However mixing it with existing actions would mean
1. Actions are managed using different APIs, it's probably not the best way to manage validation/transformation
2. Rules lack of ordering, if we are to force the such actions to be positioned in front of the list then it's effectively a separate list.

zmstone:
> I believe that introducing an entirely new message processing system outside of the rule engine would increase the overall system's complexity and the learning curve for users. I suggest expanding the capabilities of the rule engine by adding a few configuration options to allow the rule engine to terminate message flow and alter the original message content.
> 
> In my personal experience, most users do not need to terminate or modify messages in transit. However, if a user has such a requirement, they typically want all rules to have the capability to modify message flow, rather than requiring a few specific rules to refrain from altering message flow. Therefore, we can introduce a global configuration option in the rule engine's settings, something like `rule_engine.message_flow_intrusive = boolean()`, with a default value of `false`, indicating that it does not affect message flow.
> 
> With the concept of intrusive message flow, we must define the sequence of flow. We don't need to handle hooks because we've already hardcoded the order of hooks. However, the order of rules still needs to be determined. Therefore, we can add a global configuration in the rule engine: `rule_engine.message_flow_order = ["rule_name_001", "rule_name_002"]`.
> 
> For each rule, we can add a configuration, for example: `rule.rule_name_001.message_flow_action_if_matched = "stop_message_flow" | "disconnect_connection" | "none"`. This will terminate message flow or disconnect the connection after executing all actions, where "none" indicates taking no action, and the message will continue to flow to the next rule and then to other plugins if "rule_name_001" is not matched.

From UI/UX point of view, this new feature can be positioned close to the rule engine rules.
But maybe not a good idea to extend the current separation of rule + action design as I replied above
https://github.com/emqx/eip/pull/81#discussion_r1414170353

active/0024--message-validation-and-transform.md
keynslug: Very good points. A couple of questions / thoughts.

> Actions are managed using different APIs

Q: Different from which?

> Rules lack of ordering, if we are to force the such actions to be positioned in front of the list

I can very well stray away from the initial requirements, but assuming this feature is for some advanced user, then it's easy to imagine that they want to control the order as well, i.e.: 1) log and/or archive in some remote system the message completely as it is, 2) validate / transform it, 3) push through some bridge. However this starts to sound more like _pipelines_ honestly, I'm not sure if we are ready to go that far.

Anyway, I'm just exploring / expressing design ideas, please feel free to ignore them.

active/0024--message-validation-and-transform.md
keynslug: > do you mean there is no need for 'transform' vs. 'validation' ?

Mostly this. Agree that config clarity is important. I'm just trying to throw some random ideas that could help to _reduce_ the number of new concepts.

Another random thought though: if we want to _disconnect_ the client when it sends garbage, maybe it's the job for the authorization layer?

zmstone: > Another random thought though: if we want to disconnect the client when it sends garbage, maybe it's the job for the authorization layer?

Before this EIP, there were some discussions around:
1. Extend authz checks to support SQL  --- I like this idea very much ;)
2. Extend rule engine to support authz-like actions --- Product team proposed this, and I think @terry-xiaoyu had also came to this according to his comment.

This EIP is kind of striking in the middle because it's an important feature and deserves it's own space.

active/0024--message-validation-and-transform.md
zmstone: Actions have a `actions` config root, and they are so far dedicated to data integration, and managed using the `/actions` HTTP endpoint. It's would be odd if we create a `/validations` but put things under the `actions` space.

Actions can have order when pipelined and used in rule, but rules do not have order and it would be hard to define.
To support data validation as a rule, it would imply to create a singleton rule and chained (ordered actions)

keynslug: > To support data validation as a rule, it would imply to create a singleton rule and chained (ordered actions)

Do you imagine here a scenario where a customer wants that "ALL incoming messages should be validated / filtered / transformed first", before even hitting the broker (e.g. routing / hooks)? Then I agree, it sounds impossible to achieve on the data integration side alone. As a sidenote: that sounds a lot like ingress bridge + rules but for the MQTT clients themselves (i.e. not for some remote system).

However, if hitting the broker is not an issue then I don't quite understand why a singleton rule is a must. Again, a sufficiently advanced customer may have a lot of different clients with different patterns, which may require different validation / transformation applied. Thus, different pipelines.

But I agree that ordering between rules is totally unnecessary design complexity.

active/0024-message-validation.md
```diff
@@ -0,0 +1,132 @@
...
+            type = sql
+            sql = "SELECT * WHERE payload.value < 0"
+        }
+    ]
+}
+```
+
+If there are more than one validation matched for one message, all validations should be executed.
+For example, if one is configured with `topics = "t/#"` and another with `topics = "t/1"`,
+when a message is published to `t/1`, the both validations should be triggered (without ordering guarantee).
```
keynslug, line 82: Few more random thoughts (sorry).

If we are to stick to more or less the same concept of rules + actions, we may think about introducing some kind of _ingress rules_ instead, with a slightly different scope: filtering / transformation of incoming messages _before_ they hit the broker / routing / hooks. One possible action at this point is just "publish", possibly with a few knobs to tweak. For example, with rules like `SELECT 'dlq/' + topic as topic, avro_decode('schema_id', payload) as payload FROM t/# WHERE clientid like 'FW%'` we could provide both filtering, transformation and selective topic rewriting.

Similarly to the current design, neither ordering nor special handling of overlapping topic filters should be needed (supposedly). If no rules matched, the default would be to just publish message as is. I'm wondering if this could also cover the use cases for the Message Transformation EIP.

The need to disconnect clients though doesn't fit very well here I must admit, I can think of:
* Adding a knob to this "publish" action that says what to do if there's nothing to publish.
* Writing a rule like `SELECT clientid FROM '$events/ingress/message.dropped'` and attaching a "disconnect" action to it.
* Similar to the one above, but within the authz realm.

active/0024-message-validation.md
```diff
@@ -0,0 +1,132 @@
...
+            type = json # can also be 'avro' or 'protobuf'
+            schema = "my-json-schema-registration-id"
+        }
+        {
+            # SQL evaluates to empty result indicates validation falure
+            # in this example, if payload.value is less than 0, the message is considiered invalid
+            type = sql
+            sql = "SELECT * WHERE payload.value < 0"
+        }
+    ]
```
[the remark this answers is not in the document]
zmstone: it is to simplify frontend design.
SQL is like an "advanced" option.

active/0024-message-validation.md
zmstone: The journey of my iterations for this feature (partially reflected in this EIP commit history) can be summarized as a progressive shift from "Programing in SQL" towards "Programing in JSON".
I am still not quite sure if this current state is a striking a perfect balance in between, so thank you for your thoughts, it should help to stop me from drifting to far away from the SQL.

> Similarly to the current design, neither ordering nor special handling of overlapping topic filters should be needed (supposedly). If no rules matched, the default would be to just publish message as is. I'm wondering if this could also cover the use cases for the Message Transformation EIP.

Now we have to discuss this: if a rule outputs a new topic which matches  another rule ?


> Similar to the one above, but within the authz realm.
do you mean something like
`authorization.validation_on_failure = disconnect | drop` ?
this would imply that the config is globally applied to all clients, all topics and all messages ?

keynslug: > a progressive shift from "Programing in SQL" towards "Programing in JSON".

Understood. I see the point. I agree that it might overall be better UX-wise. Maybe I'm attaching too much importance to the goal of sticking to the same concepts familiar to the customers (though, arguably, they may cover more use cases rather than less). Also it seems inevitable that each new "programmable in JSON" thing will bring a set of its own concepts, slightly different from any other, which may breed complexity.

> Now we have to discuss this: if a rule outputs a new topic which matches another rule ?

Very good question. I would either lean towards the safe side: do nothing special. Or even disallow this, it was sort of an illustration of potential expressiveness anyway. Curious: what happens now in a similar situation with topic rewrite rules?

> do you mean something like
> this would imply that the config is globally applied to all clients, all topics and all messages ?

I have to admit that I don't understand the use case well enough, so that's more like a speculation. Might be something like a separate authorizer in the chain with:
```
SELECT try(avro_schema_decode(schema_id, payload)) as result, 'disconnect' as action WHERE result.error IS NOT NULL
FROM 't/topic/filter/with/avro/payloads/#'
```
Where the default action is `ignore` (i.e. when nothing was produced by the SQL expression).

---

Tangentially: disconnecting misbehaving clients sounds questionable from the resource saving perspective: it's not clear that handling consecutive reconnects is cheaper than just dropping messages.

active/0024-message-validation.md
```diff
@@ -0,0 +1,132 @@
...
+
+Each data validation rule should have a unique ID. And each rule should have a input topic-filter name which
+can be used to build a topic index for quick look-up.
+
+The common parts can be described as hocon config below:
+
+### Message Validation
+
+```
+validation1 {
```
wivwiv, line 57: Would it be better to use an array? When topics overlap, they can be prioritized according to the order of the array
```
message_validation = [
  { name = "", "topics" = ["t/#"] ... },
  { name = "", "topics" = ["t/1"] ... },
]
```

active/0024-message-validation.md
```diff
@@ -0,0 +1,132 @@
...
+    enable = true # or 'false'
+    type = validation
+    description = "drop message if it is not compabitble with my avro schema or if payload.value is less tha 0"
+    topics = "t/#" # or topics = ["t/1", "t/2"]
+    check_mode = any_pass # or all_pass
```
wivwiv, line 62: Ben suggests renaming to
```
{
  validation_strategy = "and" | "or" | "AND" | "OR"
}
```

active/0024-message-validation.md
```diff
@@ -0,0 +1,132 @@
...
+    failure_action = disconnect # (disconnect also implies 'dorp') or 'drop' to only drop the message
```
wivwiv, line 63: Discussion with Ben:

In addition to disconnect & drop message, it should allow the user to specify whether or not to output to the log:
   - Output to system log
   - The log contains basic information such as name, topic, clientid, schema, etc. No payload is needed.
   - Default log level warning

In the future, users should also be allowed to configure log message templates and set log levels.

Users can choose multiple actions, so there should be no drop message on disconnect.

The UI looks like this:
- [x] disconnect
- [ ] drop message 
- [x] output to log

active/0024-message-validation.md
```diff
@@ -0,0 +1,132 @@
...
+when a message is published to `t/1`, the both validations should be triggered (without ordering guarantee).
+
+## Configuration Changes
+
+A new config root named `message_validation` is to be added.
+The value of this config is a `Map` which has the validation ID as key, and configs as value.
+
+## APIs
+
+- GET /message_validations
```
wivwiv, line 91: Supports filtering by topic and schema：
```suggestion
- GET /message_validations?topic=foo&topic_match=t/#&schema=foo&type=json
```

active/0024-message-validation.md
```diff
@@ -0,0 +1,132 @@
...
+  To list all the message validations
+
+- PUT /message_validations
+  To update a validation
+
+- POST /message_validations
+  To create a new validation
+
+## Observerbility
```
wivwiv, line 100: Add key metrics that apply to each and all message validation policies:
- Success Rate
- Historical Success Rates: Graph of success rate metrics for the last 7 days, with support for time range choices Last hour, Last 6 hours, Last 12 hours, Last day, Last 3 days, Last 7 days
- List of clients that have failed validation
  - Client ID
  - Topic
  - Policy Name
  - First failure time
  - Last failure time
  - Failure Count
  - Successes

wivwiv (approved):

active/0024-message-validation.md
zmstone: the plan, is to index topics using the trie index.

if we want to preserve order, there would be a performance issue if for example when a thousand validations defined, but only one or two would match for each message.

active/0024-message-validation.md
zmstone: I do not have strong opinion on the name of the config, but would strongly recommend `any` `all` instead of `and` `or`.

active/0024-message-validation.md
zmstone: good idea, however not quite sure what's the purpose of `topic` in conjunction with `topic_match` by `&`.

active/0024-message-validation.md
zmstone: this is more of a generic event logging + persistence requirement to be addressed in a separate EIP.
Closing this thread.

active/0024-message-validation.md
zmstone: > Curious: what happens now in a similar situation with topic rewrite rules?

Matched only once.

> 'disconnect' as action 

Thanks, then I know it's not a global config, but a even more fine-grained control flow.

active/0024-message-validation.md
wivwiv: I think it's very similar to Rules, where through array configuration, a clear execution order can be provided to the users. 
The configuration still uses an array, and after saving, it's converted into a trie for indexing. The trie is used to find all matching rules, but the execution of the rules is still done according to the order in the array.
Is this feasible?

active/0024-message-validation.md
wivwiv: I believe "and" | "or" is more akin to logical expressions in code, making it more familiar to developers. What are the advantages of "all" | "any", and are there any other products that use it?

active/0024-message-validation.md
wivwiv: This is easy, but in the future it will also need to support user-configurable log output templates.

active/0024-message-validation.md
zmstone: not just `any` and `all`. but `any_pass` and `all_pass` 
meaning is clear to developers an non-developers too.

active/0024-message-validation.md
zmstone: ok, we may add log level as a position in the output template, but the level value itself will still deserve a separate config.

wivwiv: Is it possible to add log message templates in version 1?
And can the failure action be configured this way?
```
{
   ...
   failure_action = [
      { "type": "disconnect" },
      { "type": "drop" },
      { 
        "type": "log", 
        "level": "debug", 
        "log_format": "Client ${clientid} - topic ${topic}: message is in the wrong format and the client will be disconnected"
      }
   ]
}
```

active/0024-message-validation.md
wivwiv: You've convinced me, now I'll go convince Ben.

active/0024-message-validation.md
zmstone: not quite feasible to add log format here.
what we can do is to add a log `tag` for message validations.
then a separate feature to support dynamic log handler which can be configured to filter certain tags and customize logging fields.

active/0024--message-validation-and-transform.md
```diff
@@ -0,0 +1,162 @@
+# Message validation and transform
```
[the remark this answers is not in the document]
zmstone: Renamed the title to "Built-in Message Validation".

The decision so far is to keep the two features separate. Validation is more independent.
EIP-0025 drafts the Message Transform feature which will be kept closer to the rule-engine.

PR #82, opened 2024-01-16
feat(0025): inital draft of DS for retainer

savonarola:
[EMQX-11749](https://emqx.atlassian.net/browse/EMQX-11749)

### DIFF
--- a/active/0025-ds-for-retainer.md
+++ b/active/0025-ds-for-retainer.md
@@ -3,6 +3,7 @@
 ## Changelog
 
 * 2024-01-18: @savonarola Initial draft
+* 2024-01-22: @savonarola Add more optimizations for the "straightforward approach with optimizations"; declare it as the chosen one
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


active/0025-ds-for-retainer.md
```diff
@@ -0,0 +1,185 @@
...
+lts_index_bits | topic => #message{}
+```
+
+At the same time, we have a separate storage for the messages themselves, and we store them by topic.
+```
+topic => message
+```
+
+The storage
+* is an ordered key-value storage, where the key is a topic, and the value is the message.
```
ieQu1, line 119: There could be an alternative solution: we could add a special flag to `emqx_ds:open_db/2`, so it automatically deletes the messages from the topic when a new message is published to the topic. But timestamp of the message is still present in the key. If I am not mistaken, it shouldn't introduce too much overhead. 

This has an advantage that the iterators will be incremented when a new message is published, so it should be possible to _subscribe_ to the "retained" topics with minimal modifiction to the `emqx_persistent_session_ds` logic. It may not be particularly useful for the retainer, but this can be used for exchanging session metadata, bypassing `emqx_cm`.

The session can "subscribe" to a special topic (smth. like `$SESSION/clientid/...`) and listen, for example for takeover events. This way we can eliminate one more backchannel needed for the emqx clustering. Also, session state changes can be trivially inspected by subscribing to these special topics. 

Like I said, it's probably useless for the retainer, but if it can reuse the same logic as the session storage manager, it's a win.

ieQu1 (dismissed):

ieQu1 (approved):

PR #83, opened 2024-05-10
docs(0026): initial draft of durable shared subscription implementation

savonarola:
This is the initial draft; it describes the common idea and the session's view of the process.

We will also describe the leader manager and the leader state later.


ieQu1 (approved):

active/0026-durable-shared-subscriptions.md
```diff
@@ -0,0 +1,147 @@
...
+
+Each group is managed by a unique shared group leader process (SGL) that is cluster-wide. It has a persistent state but does not always run. It spawns when any group consumer connects and exits for some time after the last consumer disappears.
+
+An SGL is spawned by a shared group leader manager (SGLM) process. SGLM runs per node and is responsible for spawning SGLs, locating them in the cluster, and guaranteeing that there is only one SGL per group.
+
+The global registration mechanism has not yet been chosen. It may be `ra`-based, `global`-based, or even DS-specific (like registering in FoundationDB).
+
+SGL keeps track of topics belonging to the group, their streams, and stream states. It is the only entity that tracks the replay progress of these streams.
+
+The groups' consumers are persistent sessions. They connect to the SGL, and the SGL leases them streams to consume. Sessions consume these streams together with their proper streams but do not persist the progress. Instead, they report the progress to the SGL.
```
zmstone, line 29: Given that a "stream" is a super-set of the real MQTT order-ensured stream `{Clientid,PublishingTopic}`, I believe it there should be a stream 'poller' on each node for each group. The poller queries ds for messages to be distributed, and hand them out to subscriber processes, then wait for acknowledgements for qos 1 and qos 2 messages.

savonarola: The idea is that we have no "independent stream consumers" (that is, some distinguished poller).
The DS pollers are the sessions themselves; they poll DS uniformly with their own streams.

The important reason why we need a leader is to have a reliable understanding of shared subscription progress.
The other reason is the distribution of streams: who replays the concrete streams. Whithout a leader, any kind of pollers would anyway have to coordinate in order to avoid replaying the same stream.

@ieQu1 WDYT?

PR #92, opened 2024-11-13
Add 0031 for config base under cluster.hocon

zmstone:


active/0031-immutable-config-base-for-cluster.hocon
```diff
@@ -0,0 +1,73 @@
...
+- File `data/configs/cluster.hocon` (low precidence)
+  This config file is mostly hidden from the users. It holds the config changes made from the UI, API or CLI.
+
+One may argue that this is not the perfect order of overriding, but that's the fixed order in EMQX since 5.1 and we cannot change it at will. i.e. This proposal is not about changing the overriding order.
+
+The ordering itself is not the problem, but the problem is, there is lack of support for maually crafted config files under `cluster.hocon`.
+Since emqx.conf is the only option, people started putting their custom configurations in `etc/emqx.conf`, but also want to use the UI/API/CLI to change some of the configurations.
+As a result, changes made from the UI/API/CLI will override the existing merged config during the runtime, but gets overriden by the `etc/emqx.conf` when the node restarts.
+
+This is in particular a problem for emqx kubernetes operator, because it encourages the users to put all configs in one yaml block which gets mapped to `etc/emqx.conf` during the deployment.
```
id, line 32: The config in yaml is applied via API, and as a result gets mapped to cluster.hocon https://github.com/emqx/emqx-operator/blob/main/controllers/apps/v2beta1/sync_emqx_config.go#L90
It may be possible to create a custom configmap for emqx.conf, but this method is not documented https://docs.emqx.com/en/emqx-operator/latest/tasks/configure-emqx-config.html.

active/0031-immutable-config-base-for-cluster.hocon
```diff
@@ -0,0 +1,73 @@
...
+
+Make EMQX configuration files great again.
+
+## Motivation
+
+The config overriding rule since EMQX 5.1 is not easy to understand and manage.
+The current (as of 5.8) config layers are as following (in the order of overrideing precidence from high to low):
+
+- Environment variables (highest precidence)
+  Enviroment varialbes starts with `EMQX_` are translated to config keys and sits on top of the override chain.
```
id, line 17:
```suggestion
  Environment variables which start with `EMQX_` are translated to config keys and sit on top of the override chain.
```

active/0031-immutable-config-base-for-cluster.hocon
zmstone: yes, it's possible to create a config map for emqx.conf or make emqx.conf include another file,
but that's not going to help a lot if it continues to sit on TOP of cluster.hocon --- we want something to sit UNDER.

zmstone: Made an update to clarify the operator behavior

id: Clarification is still incorrect. User's yaml is NOT mapped to emqx.conf.

id: Sorry, disregard me. I need to test it again later.

PR #105, opened 2026-05-15
Add EIP 0037: EMQX feature gates

zmstone:
This EIP proposes a deployment-time feature-governance mechanism for EMQX, controlled by a single environment variable `EMQX_FEATURES`. The installer of a deployment chooses which optional features are available; the running broker and dashboard cannot enable or disable features beyond what the installer specified.

Targets release-63.

## Highlights

- **Boot-time only**; no runtime toggle.
- **One env var, two forms**: a preset (`FULL` / `ESSENTIAL`) or a comma-separated list of lowercase feature names.
- **Default `EMQX_FEATURES` unset == `FULL`** for backwards compatibility — existing deployments behave unchanged on upgrade.
- **Resolved state** held in the schema field `features.enabled`, marked read-only at runtime so the dashboard cannot mutate it.
- **Enforcement reuses the API-key scope-check layer** added on release-60. Disabled features return 403 with a distinct `FEATURE_DISABLED` code, distinguishable from `UNAUTHORIZED_ROLE` (insufficient scope).
- **Dashboard** fetches `GET /api/v5/feature_flags` and hides menu items / pages backed by disabled features.
- **Essential boot mode converges** with this proposal: `EMQX_FEATURES=ESSENTIAL` is the same shape as the separate essential-mode prototype.

## Listable optional features (13)

`dashboard` (with sub-features `dashboard.sso`, `dashboard.rbac`), `authn`, `authz`, `data_integration`, `message_transformation`, `schema_validation`, `schema_registry`, `gateways`, `cluster_link`, `multi_tenancy`, `ai`, `metrics`, `mqtt_extensions`.

Four additional features (`file_transfer`, `gcp_device`, `exhook`, `opentelemetry`) are bundled into `FULL` but not addressable individually in a custom list — see the EIP for the rationale.

## Notable design decisions

- `data_integration` is a single toggle covering the rule engine, connectors, actions, and sources — rules without actions are not a shape we want to support.
- `ai` covers both LLM completion hooks and the agent-to-agent registry under one governance switch (matches the typical compliance ask "no AI features in this deployment").
- `mqtt_extensions` is a bundle for several MQTT-protocol-adjacent features (setopts, delayed publish, topic rewrite, topic metrics, auto-subscribe, slow-subs, streams, MQ).
- The `metrics → dashboard` dependency is documented but not auto-resolved; a future EIP can decouple the Prometheus scrape endpoint onto its own listener and remove the dependency at the architecture level.
- License-driven feature intersection is explicit future work and does not change the env-var interface.

Ready for review; happy to iterate on naming, granularity, or the API-scope mapping.

### DIFF
--- a/active/0037-feature-gates.md
+++ b/active/0037-feature-gates.md
@@ -1,0 +1,567 @@
+# EMQX feature gates
+
+## Changelog
+
+* 2026-05-15: @zmstone Initial draft
+* 2026-05-15: @zmstone Map feature gates onto the API-key scope check layer
+  introduced on release-60
+* 2026-05-15: @zmstone Target release-63; compile complete feature list
+  from the release-63 `apps/` tree and resolve granularity questions
+* 2026-05-15: @zmstone Fold `psk` into basics; merge `ai_completion` and
+  `agent_registry` into a single `ai` feature
+* 2026-05-15: @zmstone Remove `file_transfer`, `gcp_device`, `exhook`,
+  `opentelemetry` from the listable vocabulary (available only when
+  `EMQX_FEATURES=FULL`); fold `telemetry` into basics (license-controlled)
+
+## Abstract
+
+This proposal introduces a deployment-time feature governance mechanism
+controlled by a single environment variable `EMQX_FEATURES`. The installer
+of an EMQX deployment chooses which optional features are available; the
+running broker and dashboard cannot enable or disable features beyond what
+the installer specified.
+
+A small set of basic components is always available: MQTT broker,
+configuration, CLI, logging, license, plugins, durable storage, audit
+log, live trace, node rebalance, retainer, TLS PSK, and outbound
+telemetry (the last is license-controlled rather than gate-controlled).
+The user-listable optional features include dashboard, authentication,
+authorization, data integration, gateways, cluster linking,
+multi-tenancy, AI features, metrics, message transformation, schema
+validation, schema registry, and MQTT extensions. A handful of further
+features (file transfer, GCP device shim, external hooks,
+OpenTelemetry) are bundled into the `FULL` preset but are not
+addressable individually in the env-var list.
+
+When all optional features are disabled, EMQX boots in *essential* mode --
+the same shape proposed independently in the boot-mode prototype. The two
+designs converge here: essential mode is just `EMQX_FEATURES=ESSENTIAL`.
+
+The mechanism is intentionally simple. Selection happens once at boot.
+There is no runtime toggle, no dependency resolution, no implicit
+enablement. Operators who need a feature must explicitly list it.
+
+## Motivation
+
+Today an EMQX installation ships with every optional feature compiled in
+and started. There is no first-class way for the operator of a regulated
+deployment to say "this installation must not expose Prometheus metrics"
+or "no rule engine is permitted here" beyond reaching into config files
+and disabling individual subsystems piecemeal -- a fragile approach
+because nothing prevents a dashboard administrator from re-enabling them.
+
+We need a single, auditable, installer-controlled switch that:
+
+* is visible in deployment manifests (k8s `env:` block, `docker run -e`,
+  systemd unit, helm `values.yaml`);
+* cannot be overridden by the runtime dashboard or REST API;
+* maps cleanly to forthcoming license-based feature gating without
+  re-shaping the interface;
+* converges with the existing essential boot-mode prototype, so the same
+  mechanism serves both "minimal footprint" and "regulated deployment"
+  use cases.
+
+## Design
+
+### Target release
+
+This proposal targets EMQX release-63. The feature list below is
+compiled from the `apps/` tree on `release-63` as of this drafting;
+features added on later branches must be added to the list (and
+classified as basic or gated) when those branches merge.
+
+### Basic components (always available)
+
+The following components are always started. They have no enable/disable
+control; turning them off would mean the broker cannot run, cannot be
+operated, or cannot meet compliance requirements that the rest of the
+product assumes.
+
+* MQTT broker (the protocol core, listeners, sessions, routing)
+* Configuration subsystem (`emqx_conf`)
+* CLI (`emqx_ctl`)
+* Logging
+* License validation (`emqx_license`)
+* Plugin loader (`emqx_plugins`; framework only -- individual plugins
+  remain opt-in via install)
+* Durable storage (`emqx_durable_storage`, `emqx_ds_backends`,
+  `emqx_ds_builtin_local`, `emqx_ds_builtin_raft`, `emqx_durable_timer`)
+* Cluster RPC (`emqx_bpapi`)
+* Audit log (`emqx_audit`)
+* Live trace (`emqx_trace`, the REST surface in `emqx_management`
+  is gated by `dashboard`)
+* Node rebalance / evacuation (`emqx_node_rebalance`,
+  `emqx_eviction_agent`)
+* Retainer (`emqx_retainer` -- can still be disabled per-listener via
+  existing MQTT config; not a feature gate)
+* TLS PSK (`emqx_psk` -- per-listener config, not a feature gate)
+* Outbound telemetry (`emqx_telemetry` -- enable/disable is determined
+  by the active license, not by feature gates)
+* Connector / bridge framework apps (`emqx_resource`,
+  `emqx_connector_aggregator`, `emqx_gen_bridge` -- loaded but idle
+  when `data_integration` is off)
+
+### Optional features (gated by `EMQX_FEATURES`)
+
+The following features can be turned off. Names are lowercase,
+underscore-separated, and form the alphabet of the explicit-list form
+described later.
+
+| Name | Apps covered | What it covers |
+| --- | --- | --- |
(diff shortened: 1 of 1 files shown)


thalesmg (commented):
I'm still at the beginning of reading this, but I could not spot mentions to "join" and "cluster".  I'm thinking that we could have problems if a cluster contains a mix of feature-limited and fully-featured nodes.  If a config change for a feature that is not present in all nodes emanates from a fully-featured node, there could be some problems with the unknown config being rejected by the other nodes.  🤔

active/0037-feature-gates.md
```diff
@@ -0,0 +1,584 @@
...
+* Node rebalance / evacuation (`emqx_node_rebalance`,
+  `emqx_eviction_agent`)
+* Retainer (`emqx_retainer` -- can still be disabled per-listener via
+  existing MQTT config; not a feature gate)
+* TLS PSK (`emqx_psk` -- per-listener config, not a feature gate)
+* Outbound telemetry (`emqx_telemetry` -- enable/disable is determined
+  by the active license, not by feature gates)
+* Connector / bridge framework apps (`emqx_resource`,
+  `emqx_connector_aggregator`, `emqx_gen_bridge` -- loaded but idle
+  when `data_integration` is off)
```
thalesmg: Would those have any use if data integrations are off?  I think the last 2 are exclusively used by integrations.  The first one could be used by more stuff such as authn/authz/cluster linking and such, so these applications would declare it as a dependency and start it independently.

active/0037-feature-gates.md
```diff
@@ -0,0 +1,584 @@
...
+
+The following features can be turned off. Names are lowercase,
+underscore-separated, and form the alphabet of the explicit-list form
+described later.
+
+| Name | Apps covered | What it covers |
+| --- | --- | --- |
+| `dashboard` | `emqx_dashboard`, `emqx_management` | Dashboard UI **and** REST API (`emqx_dashboard_sso` and `emqx_dashboard_rbac` are sub-features described below) |
+| `authn` | `emqx_auth` (chain) + backends | Authentication chain machinery |
+| `authz` | `emqx_auth` (chain) + backends | Authorization chain machinery |
```
thalesmg: this will probably require adjustments to this app, since the same app encompasses 2 distinct features.

active/0037-feature-gates.md
```diff
@@ -0,0 +1,584 @@
...
+| `data_integration` | `emqx_rule_engine`, `emqx_bridge`, `emqx_bridge_*`, `emqx_connector` | Bundle: rule engine + connectors + actions + sources |
+| `message_transformation` | `emqx_message_transformation` | Per-message transformation hooks |
+| `schema_validation` | `emqx_schema_validation` | Per-message schema validation |
+| `schema_registry` | `emqx_schema_registry` | Managed schema definitions referenced by validation and bridges |
```
thalesmg: this is not truly independent: some features already depend on this.  e.g.: message transformation, schema validation, parquet-based aggregating actions, maybe others.

active/0037-feature-gates.md
```diff
@@ -0,0 +1,584 @@
...
+* Durable storage (`emqx_durable_storage`, `emqx_ds_backends`,
+  `emqx_ds_builtin_local`, `emqx_ds_builtin_raft`, `emqx_durable_timer`)
+* Cluster RPC (`emqx_bpapi`)
+* Audit log (`emqx_audit`)
+* Live trace (`emqx_trace`, the REST surface in `emqx_management`
+  is gated by `dashboard`)
+* Node rebalance / evacuation (`emqx_node_rebalance`,
+  `emqx_eviction_agent`)
+* Retainer (`emqx_retainer` -- can still be disabled per-listener via
+  existing MQTT config; not a feature gate)
```
thalesmg: I don't think this is possible.  Or is it some very recent feature?  🤔 
AFAIK, retainer's hook is either installed or not, and runs without checking for any listener configuration.

zmstone: it's a per zone turn off of retained support (drop/reject PUBLISH).

active/0037-feature-gates.md
zmstone: ok, we can skip them.

active/0037-feature-gates.md
zmstone: ok to merge them into `auth`

active/0037-feature-gates.md
thalesmg: I don't think this is currently the case:

https://github.com/emqx/emqx/blob/1d73ffe12d43a3998469f930ad2aa09995da1432/apps/emqx_retainer/test/emqx_retainer_SUITE.erl#L1083-L1084

thalesmg: besides, there's nothing in the hook handling that mentions listener config, as mentioned.

thalesmg:
As an example: assume one node was to not start `emqx_resource` (it's in the EIP basic list, I know.  I think we can still generalize the specific example) nor any of the bridge stuff, but the other node has all features.

If the the full featured node creates a connector, then, even though there's no registered handler for this root, there could be other consequences (in this case, likely from a BPAPI call to the restricted node):

```
  {error,{{"HTTP/1.1",500,"Internal Server Error"},
          [{"date","Mon, 25 May 2026 20:13:41 GMT"},
           {"server","Cowboy"},
           {"content-length","511"},
           {"content-type","application/json"}],
          #{<<"code">> => <<"INTERNAL_ERROR">>,
            <<"message">> =>
                <<"{exception,badarg,[{ets,lookup,[emqx_resource_cache,<<\"connector:mqtt:a\">>],[{error_info,#{cause => id,module => erl_stdlib_errors}}]},{emqx_resource_cache,read,1,[{file,\"src/emqx_resource_cache.erl\"},{line,111}]},{emqx_resource_manager,lookup_cached,1,[{file,\"src/emqx_resource_manager.erl\"},{line,470}]},{emqx_connector,do_lookup,4,[{file,\"src/emqx_connector.erl\"},{line,285}]},{emqx_connector_api,v2_lookup,3,[{file,\"src/emqx_connector_api.erl\"},{line,497}]}]}">>}}}
```

So, allowing a mixed-feature cluster could be very tricky.

We'll likely need to restrict the announced bpapis depending on these features, but that might not be enough.  There are several places that indiscriminately call all nodes without checking if they support a certain bpapi...  :melting_face:

zmstone:
for cluster sync, we'll probably need to map the EMQX_FEATURES flags to a config entry like cluster config sync, the new node to join the cluster should intersect the two feature sets (`remote ∩ local`).

active/0037-feature-gates.md
```diff
@@ -0,0 +1,582 @@
...
+
+#### Dashboard sub-features
+
+When `dashboard` is enabled, two sub-toggles refine the dashboard's
+auth surface:
+
+| Sub-feature | App | What it covers |
+| --- | --- | --- |
+| `dashboard.sso` | `emqx_dashboard_sso` | Single sign-on for dashboard users |
+| `dashboard.rbac` | `emqx_dashboard_rbac` | Role-based access control for dashboard users |
```
thalesmg: I think rbac is actually a dependency of dashboard, not the other way around:

https://github.com/emqx/emqx/blob/5429aa9062b66e69b8b71002464cbafe810eefda/apps/emqx_dashboard/src/emqx_dashboard_admin.erl#L955

(and other places, such as dashboard_token and mgmt_auth)

thalesmg: might need some refactoring to change this inversion

active/0037-feature-gates.md
thalesmg: also, there are the rule sql functions that call schema registry fns and ai completion stuff.

should declaring `data_integration` without `schema_registry` and `ai` prevent booting, like it is suggested for `dashboard.sso` without `dashboard`?

active/0037-feature-gates.md
zmstone: i'd rather not over complicate it, maybe simply group all dashboard features in the dashboard feature.

active/0037-feature-gates.md
zmstone: schema registry perhaps should be removed from the flags

active/0037-feature-gates.md
savonarola: Just to note, plugins can make use of `emqx_resource`, so I'd leave them in the base setup.

hjianbo (commented):
Does the cluster support mixed nodes with different EMQX_FEATURES? If so, are there any potential risks with configuration or Mnesia replication?

active/0037-feature-gates.md
```diff
@@ -0,0 +1,582 @@
...
+its parent (`dashboard.sso` with no `dashboard`) fails boot.
+
+#### Features bundled into `FULL` but not individually listable
+
+A handful of features are part of the `FULL` preset but cannot be
+addressed by name in the explicit-list form. They are governance
+"all-or-nothing": either the deployment uses `FULL` and gets them, or
+it uses a custom list and does not.
+
+| App | Why it is not individually listable |
```
zmstone: second thought: for these apps, maybe we can make them listable after all.

they are disabled by default, but admin may wish to make them not possible to be enabled (e.g. from dashboard).

PR #106, opened 2026-05-19
docs(0036): simplify security tightening proposal

savonarola:
This PR tries to propose a simplified version of the `acl.conf` transition.

The transition of `acl.conf` involves several dependent tweaks:
* changing profile
* updating EMQX
* setting `acl.no_match=profile`
* changing `acl.conf`
 
It is not very easy to understand how they interact, and it may be even harder for an ordinary customer. 

For example, a user **must not** change `no_match` if they already do not use file ACL. If they blindly follow the upgrade instructions and set `acl.no_match=profile`, they will loosen the ACL.
Also, e.g. they may use switching file ACL on and off for testing (switching it on to allow temporary access). If they upgrade, follow the instructions, and move the permissive catch-all `{allow, all}` into `no_match=profile`, switching the file authorizer off won't restore the strict rules.

In the PR
* I tried to reduce the number of tweaks required for the upgrade: no new value for `no_match`, no need to change it.
* To make things work in a way such that ACL rules only have a chance to become _stricter_ during the process.

With the updated proposal, we are also free to introduce `acl.conf` update in 6.x


zmstone:
change the release target from 6.2 to 6.3

savonarola:
> change the release target from 6.2 to 6.3

✅

zmstone (approved):

PR #107, opened 2026-05-21
Add EIP 0038: secret registry for data-integration templates

zmstone:
Introduces a cluster-wide secret registry application that stores named byte strings, exposes a management HTTP API gated under the existing `data_integration` API-key scope, and surfaces values to consumers through three access paths matched to the shape of each consumer:

- `$secret{<name>}` placeholder for template-rendered output (HTTP bridge headers / URLs / bodies, authn / authz HTTP request templates) — log-channel-safe; the placeholder is preserved verbatim in every trace / debug path while the wire sees the substituted value.
- `get_secret('<name>')` rule SQL function as an escape hatch for cases where the secret value must participate in a SQL expression (HMAC signing, multi-input composition). Documented to leak through rule trace output when tracing is enabled — used sparingly.
- `secret://<name>` URI for HOCON secret-typed fields (action / source / connector passwords and API tokens). Slots into the existing `emqx_schema_secret` / `emqx_secret_loader` / `emqx_utils_redact` plumbing as a parallel clause to `file://`.

Targets EMQX v7.

Read-back via the management API returns metadata only — the stored byte string never leaves the broker except through one of the three resolution paths above. Backup / export excludes the registry by default.

Non-goals for v1: encryption at rest with broker-managed master key (defer to filesystem encryption), external secret-broker integration (Vault, AWS Secrets Manager), and automatic rotation / TTL.

Declined alternatives documented in the EIP: rule function as the only access surface, encryption at rest, external broker integration, returning the value via GET, TTL / auto-rotation, and allowing the placeholder in MQTT topics / payloads.

### DIFF
--- a/active/0038-secret-registry.md
+++ b/active/0038-secret-registry.md
@@ -12,6 +12,10 @@
   to a global secret only when that global secret is explicitly marked
   shareable. A global secret is tenant-invisible by default. Global→namespace
   and sibling-namespace resolution remain forbidden
+* 2026-07-24: @zmstone A stored value may be a `file://` / `env://` source
+  reference resolved at use time (per @savonarola), so file/env-delivered
+  secrets (Vault Agent, K8s secret volumes) are referenceable without copying
+  plaintext into the registry — no per-source placeholder variant
 
 ## Abstract
 
@@ -29,8 +33,12 @@ templates -- HTTP bridge headers / URLs / bodies, HTTP authn / authz
 request templates, webhook actions. It is not a general-purpose
 configuration store, not an MQTT publish payload source, and not a
 key-management service. Encryption at rest, automatic rotation, and
-secret-broker integration (Vault, AWS Secrets Manager) are explicit
-non-goals for v1.
+direct secret-broker API integration (Vault, AWS Secrets Manager) are
+explicit non-goals for v1 — though a stored value may be a
+`file://` / `env://` reference resolved at use time, so secrets
+delivered as files or environment variables (Vault Agent, Kubernetes
+secret volumes) can be referenced without copying their plaintext into
+the registry (see "Value indirection").
 
 Read-back over the API returns metadata only -- the stored byte string
 never leaves the broker except as part of a rendered template at the
@@ -108,7 +116,7 @@ This proposal targets EMQX v7.
 A new application `emqx_secret_registry` under `apps/` provides:
 
 * A cluster-wide mnesia / mria table storing
-  `#emqx_secret{{Scope, Name}, Value, ...metadata...}` rows (see
+  `#emqx_secret{key = {Scope, Name}, value = Value, ...metadata...}` rows (see
   "Storage shape"). Replicated to all nodes via the same shard
   mechanism used by existing config tables.
 * An HTTP API under `/api/v5/secrets/...` for create,
@@ -131,7 +139,7 @@ Record:
 ```
 {emqx_secret,
    key         :: {?global_ns | binary(), binary()},  %% {Scope, Name}
-   value       :: binary(),      %% raw bytes, UTF-8 well-formed
+   value       :: binary(),      %% literal bytes, or a file://|env:// source reference
    share_ns    :: boolean(),     %% global secret reachable from namespaces? default false
    description :: binary(),      %% operator-supplied free-form note, optional
    created_at  :: integer(),     %% monotonic millis
@@ -162,6 +170,10 @@ Constraints:
   cap. Names are case-sensitive.
 * `value` is at most 16 KiB. Stored as-is after a UTF-8 well-formedness
   check. PEM-encoded private keys (newline-containing) are permitted.
+  A `file://` / `env://` source reference (see "Value indirection") is
+  just a short value that passes the same check; the size / encoding
+  of the *resolved* bytes is the referenced source's concern, checked
+  at the use site like any literal value.
 * `description` is at most 512 bytes, UTF-8, control-character-free
   except whitespace.
 * The total number of secrets is capped at 1024 **per namespace**
@@ -171,6 +183,47 @@ Constraints:
   registry for everyone; the precedents are `?MAX_NUM_TNS_CONFIGS`
   (1000 managed namespaces) and `?MAX_COLLECTIONS` (512 topic-metric
   collections).
+
+### Value indirection: external sources
+
+A stored `value` is normally a literal byte string. It may instead be
+a **source reference** that EMQX resolves to bytes at use time:
+
+* `file://<path>` — read the bytes from a file on the resolving node's
+  filesystem.
+* `env://<VARNAME>` — read the bytes from an environment variable.
+
+The registry row then holds only the reference, never the underlying
+secret bytes. This is the interoperability path for operators who
+already deliver secrets through Vault Agent, Kubernetes secret
+volumes, or an init container: the external tooling drops the real
+secret as a file (or exports it into the environment), and the
+registry entry points at it. The plaintext never enters the mnesia
+table, the backup, or `cluster.hocon`, yet `$secret{<name>}`,
+`get_secret('<name>')`, and `secret://<name>` all keep working
+unchanged — resolution of the name simply yields the reference, which
+is then resolved one step further to the bytes.
+
+This reuses existing plumbing rather than adding a token per source:
+the same `emqx_secret_loader` / `emqx_schema_secret` machinery that
+already resolves `file://` for HOCON secret fields resolves the
+registry value. The source is chosen in the *value*, not pinned in the
+*template* — there is deliberately no `$secret_file{}` / `$secret_env{}`
+placeholder variant; `$secret{<name>}` is source-agnostic and the row
+decides where the bytes come from. A value is treated as a reference
+iff it matches `^(file|env)://`, the same scheme-detection convention
+`emqx_schema_secret` already uses (so, as today, a literal value that
+must itself begin with those characters is the one shape not
+expressible — not a concern for credentials).
+
+Resolution timing and failure follow the literal case exactly: the
+reference is resolved at render time (placeholders) or at connector /
+action start (`secret://` HOCON fields), and a resolution failure
+(file missing, variable unset) produces the same strict-fail as an
+unknown secret — name reported, value never. Because the stored value
+is a path or variable name and not the secret itself, it is not
+redaction-sensitive; the API's no-read-back rule is unchanged (the
+reference is still not returned, and neither is the resolved value).
 
 ### Sanitization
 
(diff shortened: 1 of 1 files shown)


active/0038-secret-registry.md
```diff
@@ -0,0 +1,525 @@
...
+constraining what shape an operator's API token may take.
+
+### Placeholder syntax
+
+A new template token `$secret{<name>}` is recognised by the renderers
+that participate in data-integration paths:
+
+* `emqx_template` (the main template renderer used by HTTP bridge
+  headers, URLs, bodies, and by `emqx_auth_http_utils`).
+* `emqx_placeholder` (the legacy placeholder substitutor used in
```
savonarola, line 270: Maybe it makes sence in v7 to 
* get rid of legacy templating
* make a more or less frormal spec for the template syntax

We have another tasks requiring templating changes, e.g. support of table name interpolation, so it would be nice to reduce adhoc changes.

zmstone: we'll probably need to support it sooner than v7.
question: would `$secret{}` be a future proof snytax ?

imanpeeman:
@zmstone what i was wondering: how about the scope of the secret? e.g. when i create it with the "global" apikey/secret I assume the secret is available system wide, when i create a secret with a namespace scoped apikey, is it then scoped to be used from the namespace only? (I need the system wide secret to be accessible from the auth http hook, but also inside the namespace for a namespace scoped http hook)

zmstone:
> is it then scoped to be used from the namespace only

not in the current proposal, i will add it as a requirement

active/0038-secret-registry.md
savonarola: > question: would $secret{} be a future proof snytax ?

I think yes.  

> support of table name interpolation

For SQL, we may need `$id{x.y}`

active/0038-secret-registry.md
```diff
@@ -0,0 +1,736 @@
...
+secret-broker integration (Vault, AWS Secrets Manager) are explicit
+non-goals for v1.
+
+Read-back over the API returns metadata only -- the stored byte string
+never leaves the broker except as part of a rendered template at the
+moment the template is used.
+
+Secrets are namespace-scoped: a secret is owned by the namespace of
+the principal that created it, a namespaced reference resolves
+against that namespace first and falls back to the global registry,
```
savonarola: Should fallback behaviour be configurable for individual global secrets? I.e. may we need secrets hidden from namespace users?

zmstone: good point. maybe add an attribute to each global secret to mark if it’s shared with namespaces, fallback only happens on shared ones

imanpeeman: I think the fallback is mainly to prevent double secret configurations. 

One of the scenario's is a multitenant target system to handle the messages. When a client connects it connects at the global auth handler. This global auth handler needs an api key to access the remote tenant, so the apiKey is stored at global level. When the device is authenticated and start delivering messages these messages are delivered in the namespace and the message forward needs also the same apikey from global to deliver the message at the target tenant. 

There might be scenario's when you indeed don't want to use the global secret, however it's mainly a configuration / use your brain when using secrets.

Maybe a configurator should pass the scope of the secret he / she wants to use so secret(::namespace::/global, secretName) then the fallback is not a resposability of the system, but of the configuration.

active/0038-secret-registry.md
```diff
@@ -0,0 +1,736 @@
...
+
+### Target release
+
+This proposal targets EMQX v7.
+
+### Component overview
+
+A new application `emqx_secret_registry` under `apps/` provides:
+
+* A cluster-wide mnesia / mria table storing `{name, value, metadata}`
```
savonarola:
```suggestion
* A cluster-wide mnesia / mria table storing `#emqx_secret{{NS, Name}, Value, ...metadata..}`
```

active/0038-secret-registry.md
```diff
@@ -0,0 +1,736 @@
...
+
+Record:
+
+```
+{emqx_secret,
+   key         :: {?global_ns | binary(), binary()},  %% {OwnerNs, Name}
+   value       :: binary(),      %% raw bytes, UTF-8 well-formed
+   description :: binary(),      %% operator-supplied free-form note, optional
+   created_at  :: integer(),     %% monotonic millis
+   updated_at  :: integer()      %% monotonic millis
```
savonarola:
```suggestion
   updated_at  :: integer(),     %% monotonic millis
   extra :: #{}
```
?

active/0038-secret-registry.md
```diff
@@ -0,0 +1,736 @@
...
+  `ns:<namespace>::<role>` and parsed by
+  `emqx_dashboard_rbac:parse_role/1`.
+* At the HTTP boundary `emqx_dashboard:get_namespace/1` reads it out
+  of `auth_meta`, defaulting to `?global_ns`.
+* A `resolve_namespace/2` minirest filter stores the effective
+  namespace as `resolved_ns` on the request, applying the established
+  rule: a **global** admin may act on any namespace by passing
+  `?ns=<namespace>`; a **namespaced** admin is pinned to their own and
+  gets `not_authorized` if they pass a different one. This mirrors
+  `emqx_connector_api:parse_namespace/1`.
```
savonarola, line 561: Maybe no need to pin the detailed implementation since we already have a well established mechanism governed outside the scope of the EIP.

active/0038-secret-registry.md
```diff
@@ -0,0 +1,736 @@
...
+inherited default — and it is safe because the only configurations
+that resolve through `NS` are configurations *owned by* `NS`. A
+tenant can redirect their own bridges to their own credential; they
+cannot affect anyone else's.
+
+For the cases where an operator wants to defeat shadowing and pin a
+reference to the global registry explicitly, both reference syntaxes
+accept a qualified form:
+
+* `$secret{global::<name>}` for the placeholder.
```
savonarola: maybe 
```suggestion
* `$secret{global.<name>}` for the placeholder.
```
to keep lgically in sync with regular templating. Like in nested bindings, we have a nested virtual collections of secrets

active/0038-secret-registry.md
```diff
@@ -0,0 +1,736 @@
...
+* `secret://global/<name>` for the HOCON URI.
+
+The unqualified forms keep their fallback semantics. `global` is
+already a reserved namespace name
+(`emqx:is_reserved_namespace/1` rejects `global`, `undefined`,
+`null`, `none`), so the qualifier cannot collide with a real
+namespace.
```
savonarola: Maybe a bit unncecessary/misleading clarification since 
* `name` matches `^[A-Za-z0-9][A-Za-z0-9_-]{0,62}$`;
* `$secret{<ns>::<name>}` is not allowed anyway
?

active/0038-secret-registry.md
```diff
@@ -0,0 +1,904 @@
...
+`cluster.hocon`, visible via the dashboard, included in `emqx ctl data
+export` artifacts, and surfaceable in any trace or debug log that
+renders the template.
+
+Three concrete pain points follow:
+
+* **Operational**: rotation is N-touch. Tooling to mass-update the
+  same value across configs is missing today and would be brittle if
+  added (matching by string equality across heterogeneous templates).
+* **Audit / governance**: an operator who needs to configure a new
```
savonarola: This is probably mostly a foolproof measure since an operator can know out the token inserting it into the configured requests

active/0038-secret-registry.md
```diff
@@ -0,0 +1,904 @@
...
+namespaced admin must not be able to enumerate another tenant's
+secret *names*, nor the names of **unshared** global secrets, since
+names are chosen by operators and routinely describe the system they
+unlock. The secrets API therefore filters list / get results **in the
+handler**, not relying on the RBAC layer. For a caller whose
+`resolved_ns` is a namespace `NS`, a row is visible iff it is owned by
+`NS`, **or** it is owned by `?global_ns` and has
+`shared_ns = true`. Everything else — other namespaces' secrets and
+unshared global secrets — is omitted from lists and returns `404` on
+direct `GET`.
```
savonarola: IMO the overpermissive `do_check_rbac` is a much greater concern than listing secret names in another namespace 🤔 Maybe an overcomplication?

active/0038-secret-registry.md
```diff
@@ -0,0 +1,904 @@
...
+support fetching secrets from external secret brokers at render
+time. Rejected for v1 on two grounds: it expands the surface area
+significantly (broker plugins, credential management for the broker
+connection, timeout / cache semantics for slow brokers, error
+handling for "broker reachable but secret missing" vs "broker
+unreachable"), and it shifts the operational model from "EMQX is
+self-contained" to "EMQX has a hard runtime dependency on an
+external secret service." Operators who want Vault-backed secrets
+can run a Vault Agent sidecar that templates a HOCON file containing
+`emqx_secret_registry` CRUD calls and re-applies on rotation.
```
savonarola: The new secret store is still functionally somewhat like a small in-house Vault, Kubernetes Secrets, etc. I worry that this may frustrate users who already use Vault or Kubernetes Secrets.

Previously, secret management was outside EMQX’s scope, so:

* Some ad hoc solutions were emotionally acceptable.
* Vault/Kubernetes secrets could still be injected into configs in a common way e.g. through environment variables.

Now we offer a _brand-new_ tool, but it lacks even the old level of interoperability, since values are stored in the database (not a config).

Maybe we should still consider implementing some basic interoperability with environment variables and secret volumes. For example, a secret value could be an explicit source URI:

* env://EMQX_SECRET_MY_NAME
* file:///some/configured/secret/volume/path/my_name

EMQX would store the source URI in the registry and resolve the actual value at use time. This would allow `$secret{my_name}` to use secrets delivered by Vault, Kubernetes, or another external system without copying their values into the EMQX database/sidecars/templating/etc.

zmstone: you mean like `$secret_file{name}` and `$secret_env{name}` ?

active/0038-secret-registry.md
zmstone: if `.` means value path in a object/struct, i'd keep `::` for namespace.

active/0038-secret-registry.md
savonarola: I meant just special values in the value field of the database
```
#emqx_secret{
  key =  {global, <<"my_name">>},
  value = <<"file:///some/configured/secret/volume/path/my_name">>
}
```
or
```
#emqx_secret{
  key =  {global, <<"my_name">>},
  value = {file, <<"/some/configured/secret/volume/path/my_name">>}
}
```

IMO pinning exact source in the template (`$secret_file{}`/`$secret_env{}`) is not a necessary limitation

active/0038-secret-registry.md
```diff
@@ -0,0 +1,946 @@
...
+### Target release
+
+This proposal targets EMQX v7.
+
+### Component overview
+
+A new application `emqx_secret_registry` under `apps/` provides:
+
+* A cluster-wide mnesia / mria table storing
+  `#emqx_secret{{Scope, Name}, Value, ...metadata...}` rows (see
```
savonarola: I made a typo in the previous suggestion 🙈 
```suggestion
  `#emqx_secret{key = {Scope, Name}, value = Value, ...metadata...}` rows (see
```

savonarola (approved):

PR #108, opened 2026-05-26
doc(0036): use deny instead of nomatch for external backend failures

savonarola:
This may be an overkill.
However, current behaviour may be vulnerable as soon as anyone uses an external backend for blacklist logic.


zmstone (approved):
the chain on-error fallback is now a well established concept now.
we must announce this as a breaking change.

PR #110, opened 2026-06-02
Add EIP 0039: session buffer byte observability

hjianbo:
This PR adds an EIP to discuss lightweight observability for payload bytes retained by MQTT memory-session buffers.

The initial proposal covers:

- one per-session `total_payload_bytes` field, calculated from MQTT payload bytes retained by memory-session mqueue and inflight state
- a throttled warning log controlled by `sysmon.session.buffered_payload_high_watermark` (`0` disables the warning log)
- an on-demand cluster `emqx ctl session-top` CLI that scans cached session data and writes CSV output
- implementation boundaries for cached session data and paced top-K scans
- open questions for future aggregate metrics/alarms and a possible byte-volume limit

The initial proposal intentionally does not include node stats, Prometheus gauges, or EMQX alarms. Those are documented as deferred alternatives for later discussion.

Checks run locally:

- `git diff --check`

### DIFF
--- a/active/0039-session-buffer-byte-observability.md
+++ b/active/0039-session-buffer-byte-observability.md
@@ -1,0 +1,233 @@
+# Session Buffer Byte Observability
+
+## Changelog
+
+* 2026-06-04: @hjianbo Clarify missing configuration behavior during upgrades.
+* 2026-06-02: @hjianbo Initial draft.
+
+## Abstract
+
+This proposal adds lightweight observability for payload bytes retained by MQTT
+memory-session buffers.
+
+Today EMQX limits session buffering mainly by message count
+(`mqtt.max_mqueue_len`).  A small number of large payloads can still retain a
+large amount of resident memory in the session mqueue and inflight window.  This
+proposal exposes that payload volume per session, adds throttled warning logs,
+and provides an on-demand cluster top-K CLI for diagnosis.
+
+## Motivation
+
+Large MQTT payloads are stored as Erlang refc binaries.  These referenced
+binaries are not counted as ordinary heap memory of the session process, so
+existing heap-size based `force_shutdown` policies cannot protect a node from
+this specific memory-retention pattern.
+
+The risk becomes more visible during reconnect storms or session takeover.
+A session holding hundreds of MB, or even GB, of buffered payloads may
+cause a sharp VM memory spike when its state is transferred to a new owner.
+Many concurrent takeovers can further increase Erlang distribution pressure
+and make the node harder to diagnose.
+
+## Design
+
+### User-facing data
+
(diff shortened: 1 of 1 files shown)


active/0039-session-buffer-byte-observability.md
```diff
@@ -0,0 +1,291 @@
...
+
+Expose node-level totals in `/stats`:
+
+* `sessions_mqueue_bytes`
+* `sessions_inflight_bytes`
+* `sessions_buffered_bytes`
+
+Durable session backlogs that are stored outside the channel process are not
+resident session-buffer memory.  They report `0` for these fields unless a
+future design adds equivalent resident-byte accounting for them.
```
ieQu1: There's no need.

active/0039-session-buffer-byte-observability.md
```diff
@@ -0,0 +1,290 @@
...
+  `buffered_bytes`.
+
+## Implementation
+
+### Message size algorithm
+
+Use one stable message-size estimator for both mqueue and inflight accounting:
+
+```erlang
+message_bytes(Msg) = erlang:external_size(Msg).
```
zmstone: don't do this, byte_size(Payload) is enough.

active/0039-session-buffer-byte-observability.md
```diff
@@ -0,0 +1,290 @@
...
+Add `sysmon.session_buffer` to control sampling and alarms:
+
+```hocon
+sysmon.session_buffer {
+  enable = true
+  sample_interval = 1m
+  node_high_watermark = 1GB
+  client_high_watermark = 100MB
+  alarm_sustain_duration = 15s
+  top_clients_count = 10
```
zmstone: i thought 'top' is declined.

zmstone (commented):
propose to only add per-client observability in this EIP. 

Only inc/dec byte_size(Payload) and keep the gauge in channel info (cached in ets).

no aggregation,
no alarms

gauge aggregation/top-K can be implemented separately e.g. as a plugin.

active/0039-session-buffer-byte-observability.md
```diff
@@ -0,0 +1,225 @@
...
+* keep a bounded local top K instead of materializing all sessions;
+* return only local top K rows to the initiating node;
+* merge partial results on the initiating node and write the final CSV.
+
+For `N` local sessions and requested `K`, local memory usage should be `O(K)` and
+scan CPU should be `O(N log K)` or better.
+
+## Backwards Compatibility
+
+N/A
```
zmstone: consider hot-upgrade

- config read at runtime should handle the case when `undefined` is returned (do not assume config default value is always populated)

zmstone (approved):

PR #112, opened 2026-06-10
docs(0035): qos0 always enter mqueue

zmstone:


zmstone:
declined because this may cause qos0 messages to be dropped more than necessary.
e.g. if tcp send buffer is large enough to hold more than mqueue can hold,
enqueue before send may cause send buffer starvation.

PR #113, opened 2026-07-02
doc(0040): trusted client attributes

savonarola:
This EIP introduces a provenance-aware security context for `ClientInfo` users.
Instead of treating **full** runtime `ClientInfo` as authority-bearing data, post-authentication security-sensitive consumers receive a **trusted projection** derived only from EMQX-controlled fields, authenticator outputs, and explicitly validated client fields.


active/0040-trusted-client-info.md
```diff
@@ -0,0 +1,267 @@
...
+
+To:
+
+![New model](0040-assets/attr-boundary.svg)
+
+* We explicitly treat some `clientinfo` attributes as trusted, and some not.
+* The trusted attributes are those that come from trusted sources.
+  * We trust, e.g., listener/zone attributes because they are set by EMQX itself.
+  * We trust attributes that come from authenticator, like `auth_expire_at`, `acl`, `is_superuser`, any additional attributes set by the authenticator.
+  * We trust attributes that were validated by the authenticator, e.g. `clientid`, `tns` if clientid-based builtin authentication was used.
```
zmstone, line 75: you mean by parsing the auth-chain configuration?
for external authenticator, we parse the request templates to find placeholders?

active/0040-trusted-client-info.md
```diff
@@ -0,0 +1,267 @@
...
+It has the backward compatible simple shape, but with untrusted fields stripped out.
+
+To strip out untrusted fields from `clientinfo` we do the following:
+
+* We have a static list of trusted fields that are always included in the output, like `listener` or `zone`.
+* We move authenticator-originated fields into the trusted submap:
+  ```erlang
+  #{
+      listener => ...,
+      clientid => ...,
```
zmstone, line 96: If authn is username based, maybe clientid is not checked hence cannot be trusted?
Though to be perfectly secure, one should never skip checking client ID -- so if we assume client ID is always checked, then we can indeed treat it safe.

zmstone (commented):
~~maybe build~~

active/0040-trusted-client-info.md
savonarola: for external authenticators, we of course cannot guess the used templates, so they will need to return trusted attributes (`clientinfo` mask) explicitly.

active/0040-trusted-client-info.md
savonarola: This is the "original" full clientinfo. The example of a stripped one goes further.

active/0040-trusted-client-info.md
zmstone: this can be easy to support for HTTP authenticator.
but will need to define the contract for other backends

active/0040-trusted-client-info.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+
+And deduce part of the mask from authentication metadata (used vars in the templates). However, normally there should not be many data in the mask, mainly `clientid`/`username` and maybe `tns`.
+
+## Configuration Changes
+
+We will need:
+* Tweaks to explicitly loosen the trust requirements where trusted client info is applied.
+* Tweaks to explicitly mark some attributes as trusted.
```
zmstone: maybe add some examples

zmstone (dismissed):

zmstone (approved):

PR #119, opened 2026-08-12
EIP-0044: Encrypted and authenticated data backups

zmstone:
Adds EIP-0044 proposing optional authenticated encryption for EMQX data backups.

Backups are plaintext today and import checks only edition/version, so restore both discloses all cluster secrets (an archive holds `cluster.hocon` + every Mnesia dump) and injects arbitrary config/DB state.

The proposal encrypts a backup with an AEAD cipher under an operator-supplied `file://` key:
- **Confidentiality + integrity + provenance** in one primitive (AEAD).
- **Producing cluster stores no key** — the create key is ad-hoc, supplied per request/CLI and discarded.
- **Non-forgeable API import** — the decrypt key must be provisioned out of band via `ctl`, never carried in the import request; CLI import is the trusted, exempt plane.
- **Node policy** `require_encryption` (hardened default), and the archive's self-reported version never relaxes the check (blocks forge-as-old downgrade).
- Decrypt-before-extract, envelope key with rotation, audit of import attempts.

Declined alternatives capture why: on-cluster signing key (secret on every node), asymmetric signing (deferred to the cross-org case), key-in-import-request (attacker supplies both), version-based relaxation (downgrade), non-AEAD encryption (no integrity).

Next free number: main is at 0042, #117 claims 0043.

### DIFF
--- a/active/0044-backup-encryption.md
+++ b/active/0044-backup-encryption.md
@@ -1,0 +1,230 @@
+# Encrypted and Authenticated Data Backups
+
+## Changelog
+
+* 2026-08-12: @zmstone Initial draft
+
+## Abstract
+
+EMQX data backup archives are plaintext `tar.gz` files, and the import path
+applies them after checking only the EMQX edition and version. An archive
+carries the whole cluster state: `cluster.hocon` and every Mnesia table dump,
+which include API keys, authentication and authorization records, connector
+passwords, and other secrets. Restore therefore both discloses those secrets to
+anyone who reads a backup file and injects arbitrary configuration and database
+state into the cluster when a file is imported.
+
+This EIP adds optional authenticated encryption for backups. A backup is
+encrypted with an operator-supplied key using an AEAD cipher, so the archive is
+confidential and tamper-evident. Import through the REST API succeeds only when
+the matching key was provisioned out of band, so an API caller cannot import an
+archive the operator did not authorize. The producing cluster stores no key.
+
+## Motivation
+
+Restore is one of the most powerful operations in the product. It rewrites
+`cluster.hocon`, which can enable plugins, change `os_mon` limits, and install
+rule-engine actions, and it loads Mnesia table dumps that include credentials.
+An attacker who gets a crafted archive imported gains control of the cluster.
+The current import path does not check integrity or provenance, so a tampered or
+attacker-supplied archive is applied as-is once a caller reaches the import
+endpoint.
+
+Backups are also a confidentiality problem on their own. A plaintext archive
+that leaves the node, for example copied to object storage or shared for
+support, exposes every cluster secret to anyone who reads the file.
+
+Two properties are needed:
+
+* Confidentiality of the archive at rest and in transit.
+* A guarantee that an archive imported through the REST API was produced by a
+  party the operator trusts and was not modified.
+
+Restore is a dangerous capability that should be gated, in the same spirit as
+the control over which plugin CLI commands are allowed to run.
+
+## Design
+
+### Overview
+
+A backup is encrypted with an AEAD cipher (for example AES-256-GCM or
+XChaCha20-Poly1305). AEAD gives confidentiality and integrity in one step: an
+archive that decrypts and authenticates under the expected key is proven both
+unmodified and produced by a holder of that key. A separate signature or MAC is
+not required.
+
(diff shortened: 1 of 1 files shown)


active/0044-backup-encryption.md
```diff
@@ -0,0 +1,230 @@
...
+extracting members from an unauthenticated archive would mean trusting
+attacker-controlled content to make a security decision, and would expose the
+extractor to malicious archive members. Decryption is the outer gate; the
+existing edition, version, and namespace handling runs on the authenticated
+plaintext.
+
+### Node policy
+
+Whether encryption is required is a property of the importing node, not of the
+archive. A node setting `data_backup.import.require_encryption` controls it, and
```
zmstone: there is no data_backup config root.

`node.data_import_encryption = per_profile | true | false.`

active/0044-backup-encryption.md
```diff
@@ -0,0 +1,296 @@
...
+  key may be passed as a `file://` argument. CLI import is the trusted plane.
+
+Moving a backup between clusters is therefore: pick a key, create the backup on
+the source with that key, provision the key on the importer through `ctl`, then
+import. The key is exchanged once per trust relationship, not once per backup.
+
+### Envelope encryption
+
+Each backup is encrypted with a fresh random data key. The data key is wrapped
+with the operator-supplied key and stored, with a key identifier and the AEAD
```
savonarola: NIT: a bit difficult to read: in the "Key handling" section, it is not clear that the key in question is not an actual data encryption key. Maybe introduce explicit terms like KEK (key encryption key) and DEK (data encryption key) to avoid confusion when just "the key" is mentioned.

active/0044-backup-encryption.md
```diff
@@ -0,0 +1,296 @@
...
+nonce, in a small cleartext header prepended to the archive. Import reads the key
+identifier, selects the matching provisioned key, unwraps the data key, and
+decrypts. Envelope encryption lets the operator rotate the provisioned key
+without re-encrypting existing archives and lets the importer reject an archive
+whose key it does not hold before doing any other work.
+
+### Order of operations on import
+
+Import decrypts and authenticates the archive first, from the raw uploaded bytes,
```
savonarola, line 123: Should it decrypt the key first?

active/0044-backup-encryption.md
```diff
@@ -0,0 +1,296 @@
...
+that bypasses the requirement is recorded as such. The create-backup key is a
+`file://` reference and is never recorded.
+
+## Configuration Changes
+
+* `node.data_import_encryption`: enum `per_profile | true | false`. Default
+  `per_profile` (required under the hardened profile, not required under the legacy
+  profile). When encryption is required, REST API import requires an encrypted
+  archive that authenticates under a provisioned key.
+* Create-backup accepts an optional key. The REST API takes it in the
```
savonarola, line 215: Should plaintext backups be also limited by policy?

active/0044-backup-encryption.md
```diff
@@ -0,0 +1,296 @@
...
+
+Encryption wraps a finished archive. EMQX builds the plaintext `tar.gz` first,
+exactly as today, and then encrypts the whole file. The encrypted output is not a
+`tar` and does not carry members of its own; it is a single opaque blob with a
+small cleartext header.
+
+* Plaintext backup: `<name>.tar.gz`, unchanged.
+* Encrypted backup: `<name>.tar.gz.enc`.
+
+The `.enc` file is `header || AEAD(plaintext tar.gz)`. The header is cleartext and
```
savonarola, line 141: maybe describe in detail the header fields

active/0044-backup-encryption.md
zmstone: good point.

active/0044-backup-encryption.md
zmstone: hmm, the key IS the data encryption key.
but agree, an explicit term will help.

active/0044-backup-encryption.md
zmstone: it's just the key identifier as plaintext prepended to the enc file

active/0044-backup-encryption.md
zmstone: ~header is not structured data, it's the key-id~
sorry, see wire format described below

